PR #1, opened 2022-04-28
Use new cirq transformer API for quantum_chess

dstrain115:
- Convert CircuitTransformer to use call to conform to Transformers
API.
 - Convert SycamoreDecomposer to a free function that conforms to
Transformers API.

Continuation of ReCirq #276

### DIFF
--- a/unitary/quantum_chess/circuit_transformer.py
+++ b/unitary/quantum_chess/circuit_transformer.py
@@ -272,7 +272,6 @@ class ConnectivityHeuristicCircuitTransformer:
                 if len(op.qubits) == 1:
                     sq.add(op.qubits[0])
                 else:
-                    print(op)
                     q1, q2 = op.qubits
                     if q1 not in g:
                         g[q1] = []
@@ -430,13 +429,14 @@ def decompose_into_sqrt_iswap(
     and control-X gates with multiple controls (TOFFOLI gates).:w
     """
     if context is None:
-        context = cirq.TransformerContext()
+        context = cirq.TransformerContext(deep=True)
     converted_iswap_circuit = cirq.map_operations_and_unroll(
         circuit.unfreeze(False),
         transform_controlled_iswap,
-        deep=True,
+        deep=context.deep,
         tags_to_ignore=context.tags_to_ignore,
     )
     return cirq.optimize_for_target_gateset(
-        converted_iswap_circuit, gateset=SQRT_ISWAP_GATESET
+        converted_iswap_circuit, gateset=SQRT_ISWAP_GATESET,
+        context=context
     )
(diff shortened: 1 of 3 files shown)


unitary/quantum_chess/circuit_transformer.py
```diff
@@ -400,48 +400,43 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
         return cirq.Circuit(updater.add_swaps())
 
 
-class SycamoreDecomposer(cirq.PointOptimizer):
-    """Optimizer that decomposes all three qubit operations into
+def transform_controlled_iswap(op: cirq.Operation, index: int) -> cirq.OP_TREE:
```
tanujkhattar, line 403: The implementation of `controlled_iswap` decompositions also use deprecated optimizers -- will these be updated in a follow-up PR? 

https://github.com/quantumlib/unitary/blob/5d7e70966dd0061c3473bb29984a38a3cb32c943/unitary/quantum_chess/controlled_iswap.py#L119


Also, `cirq.optimize_for_target_gateset(circuit, gateset=cirq.SqrtIswapTargetGateset(), ignore_failures=False)` is able to convert the controlled iswap cases as well (see the [colab](https://colab.sandbox.google.com/drive/1sjacBgm1oqa2giTQ_QUlu_GxKqJ9GLsR?resourcekey=0-UIKWzHrpTw58dWE-VPnqlw&usp=sharing#scrollTo=E3sT1gth8xJR)). My guess is that the custom decompositions used below are more efficient than the default ones?

unitary/quantum_chess/circuit_transformer.py
```diff
@@ -400,48 +400,43 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
-        if new_ops:
-            return cirq.PointOptimizationSummary(
-                clear_span=1, clear_qubits=op.qubits, new_operations=new_ops
-            )
+    if context is None:
+        context = cirq.TransformerContext()
+    converted_iswap_circuit = cirq.map_operations_and_unroll(
+        circuit.unfreeze(False),
+        transform_controlled_iswap,
+        deep=True,
```
tanujkhattar, line 437: Let's change this to `context.deep` and create the default context above as `cirq.TransformerContext(deep=True)` ?

unitary/quantum_chess/circuit_transformer.py
```diff
@@ -400,48 +400,43 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        tags_to_ignore=context.tags_to_ignore,
+    )
+    return cirq.optimize_for_target_gateset(
+        converted_iswap_circuit, gateset=SQRT_ISWAP_GATESET
+    )
```
tanujkhattar, line 442: Pass the `context=context` so that `deep=True` is propagated in all the nested transformers as well.

tanujkhattar (changes requested):

tanujkhattar (approved):
LGTM % comment regarding using deprecated optimizers in `controlled_iswap` implementation.

PR #3, opened 2022-05-05
Update some deprecated functions

dstrain115:
- device.qubit_set() is now device.metadata.qubit_set

PR #4, opened 2022-05-05
Use networkx graph for computing adjacency

dstrain115:
- As mentioned in #3, use graph_nx of the device
to construct adjacency dictionary.

### DIFF
--- a/unitary/quantum_chess/initial_mapping_utils.py
+++ b/unitary/quantum_chess/initial_mapping_utils.py
@@ -28,7 +28,6 @@ def build_physical_qubits_graph(
     Args:
       device: The device from which to build a physical qubits graph.
     """
-    g = defaultdict(list)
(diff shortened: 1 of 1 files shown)


unitary/quantum_chess/initial_mapping_utils.py
```diff
@@ -29,10 +29,8 @@ def build_physical_qubits_graph(
       device: The device from which to build a physical qubits graph.
     """
     g = defaultdict(list)
```
tanujkhattar, line 31: Remove unused variable.

tanujkhattar (approved):

unitary/quantum_chess/initial_mapping_utils.py
dstrain115: Done.

PR #6, opened 2022-05-11
Adds qudit gates to unitary API

dstrain115:
- Adds SwapPow and ISwapPow Gates for Qudits
- By default, this will do a swap and iswap gate for qudits.
- An exponent can be set to change this to do sqrt iswap and sqrt swap.

### DIFF
--- a/unitary/alpha/qudit_gates.py
+++ b/unitary/alpha/qudit_gates.py
@@ -32,3 +32,59 @@ class QuditPlusGate(cirq.Gate):
 
     def _circuit_diagram_info_(self, args):
         return f"[+{self.addend}]"
+
+
+class QuditSwapPowGate(cirq.Gate):
+    def __init__(self, dimension: int, exponent: int = 1):
+        self.dimension = dimension
+        self.exponent = exponent
+
+    def _qid_shape_(self):
+        return (self.dimension, self.dimension)
+
+    def _unitary_(self):
+        size = self.dimension * self.dimension
+        arr = np.zeros((size, size), dtype=np.complex64)
+        for x in range(self.dimension):
+            for y in range(self.dimension):
+                g = np.exp(1j * np.pi * self.exponent / 2)
+                coeff = -1j * g * np.sin(np.pi * self.exponent / 2)
+                diag = g * np.cos(np.pi * self.exponent / 2)
+                arr[x * self.dimension + y, y * self.dimension + x] = coeff
+                arr[x * self.dimension + y, x * self.dimension + y] = diag
+        return arr
+
+    def _circuit_diagram_info_(self, args):
+        if not args.use_unicode_characters:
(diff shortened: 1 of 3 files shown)


unitary/alpha/qudit_gates.py
```diff
@@ -32,3 +32,59 @@ def _unitary_(self):
...
+
+class QuditSwapPowGate(cirq.Gate):
+    def __init__(self, dimension: int, exponent: int = 1):
+        self.dimension = dimension
+        self.exponent = exponent
+
+    def _qid_shape_(self):
+        return (self.dimension, self.dimension)
+
+    def _unitary_(self):
```
losos0, line 45: The matrix is not unitary. Some columns are all zero

unitary/alpha/qudit_gates.py
```diff
@@ -32,3 +32,59 @@ def _unitary_(self):
...
+
+class QuditISwapPowGate(cirq.Gate):
+    def __init__(self, dimension: int, exponent: int = 1):
+        self.dimension = dimension
+        self.exponent = exponent
+
+    def _qid_shape_(self):
+        return (self.dimension, self.dimension)
+
+    def _unitary_(self):
```
losos0, line 75: Same problem here

unitary/alpha/qudit_gates.py
dstrain115: Thanks.  I was worried this matrix was not correct.  I will take another stab at it tomorrow.

dstrain115: Partially fixed these gates and added a test to check unitarity for the gates.  iSwap is still not quite right though.  (and I don't have high confidence in swap yet either).  One step closer though.

unitary/alpha/qudit_gates.py
```diff
@@ -16,7 +16,43 @@
 import cirq
 
 
+class QuditXGate(cirq.Gate):
+    """Performs a X_0a gate.
```
everthemore, line 20: Would it make sense to have a general X_{ab} gate instead (where a is not always 0, I mean)?

dstrain115: Done.

unitary/alpha/qudit_gates.py
```diff
@@ -16,7 +16,48 @@
...
+        arr = np.zeros((self.dimension, self.dimension))
+        arr[self.source_state, self.destination_state] = 1
+        arr[self.destination_state, self.source_state] = 1
+        for i in range(self.dimension):
+            if i != self.source_state and i != self.destination_state:
+                arr[i, i] = 1
+        return arr
+
+    def _circuit_diagram_info_(self, args):
+        return f"X(0_{self.state})"
```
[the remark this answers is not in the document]
dstrain115: Done.

PR #11, opened 2022-05-18
Add Move and PhasedMove effects.

bartmate:
I added the effect which corresponds to a move in quantum chess.

### DIFF
--- a/unitary/alpha/qubit_effects_test.py
+++ b/unitary/alpha/qubit_effects_test.py
@@ -40,6 +40,8 @@ def test_move():
     piece2 = alpha.QuantumObject("b", 0)
     board.add_object(piece1)
     board.add_object(piece2)
+    results = board.peek([piece1, piece2], count =100)
+    assert all(result == [1,0] for result in results)    
     alpha.Move()(piece1, piece2)
     expected_circuit = cirq.Circuit()
     a = cirq.NamedQubit("a")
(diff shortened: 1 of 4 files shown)


unitary/alpha/qubit_effects_test.py
```diff
@@ -34,6 +34,36 @@ def test_superposition():
...
+    piece2 = alpha.QuantumObject("b", 0)
+    board.add_object(piece1)
+    board.add_object(piece2)
+    alpha.Move()(piece1, piece2)
+    expected_circuit = cirq.Circuit()
+    a = cirq.NamedQubit("a")
+    b = cirq.NamedQubit("b")
+    expected_circuit.append(cirq.X(a))
+    expected_circuit.append(cirq.SWAP(a, b))
+    assert board.circuit == expected_circuit
```
dstrain115, line 49: Can you also test expected results too?  I hit this when my circuit was wrong before.

For instance,

```
results = board.peek([piece1, piece2], count =100)
assert all(result == [1,0] for result in results)
```

Same for below.

dstrain115 (approved):
Looks good.  Recommend extending the test slightly.

Thanks for contributing and congratulations on being the first new contributor!!!

unitary/alpha/qubit_effects_test.py
bartmate: I added two tests of `results` for both operations: One before the operation and one after the operation.

bartmate: What is the next step?

dstrain115: Next step is to merge it into main, which I did for you.  Thanks for contributing!

PR #24, opened 2022-06-07
More complete Quantum TicTacToe implementation

dstrain115:
- Fix TicTacSplit to use sqrtSWAP(01) [or 02] instead
of sqrtSWAP of all states
- Measurements are implemented.
- Simple ASCII board to try it out.


everthemore (approved):
Thank you for doing all of the heavy lifting on this 👍

Issue #22, opened 2022-06-06
Unitary: Duplicate named quantum objects are allowed but fail later


dstrain115:
Accidentally I added a qobject to the QWorld with a name which already existed in another qobject I added to the world earlier. It worked without error. I got an error later, when I tried to use it in a peek. It seems that cirq uses the name as an identifier. 

Here, I would suggest some kind of check when the object is added to the world.

PR #25, opened 2022-06-07
raise ValueError if duplicate objects added

dstrain115:
Fixes: #22


dstrain115:
Fixes the issue that @bartmate mentioned.

cognigami (approved):

Issue #17, opened 2022-05-31
Outdated version of ReCirq code


losos0:
The initial copy is missing, at least, https://github.com/quantumlib/ReCirq/pull/278.

dstrain115:
It looks like that PR was copied over, but it was inadvertently squashed by this commit:
https://github.com/quantumlib/unitary/commit/bbfcfc1159e9dedfdb047f191683c5714f22fda2

dstrain115:
I looked at the last 5 or 6 commits in quantum_chess in ReCirq, and I think this is the only commit missing.

@losos0 What do you mean by a history rewrite?  I think the commit history has the history from ReCirq.

losos0:
Oh yes it's there, my bad.

PR #26, opened 2022-06-07
Update quantum chess code

dstrain115:
Re-add quantum chess commit #278 which was inadvertently squashed during import.

Fixes: #17


losos0:
LGTM

PR #27, opened 2022-06-07
Add ASCII board for Quantum TicTacToe board

dstrain115:
- Add preliminary ASCII board for Quantum TicTacToe
- Displays probability for blank (.) and X and O.


losos0:
Alternative display idea: fill the 3x5 (or whatever) area of the cell with `X`, ` `, or `O` in proportional to the probabilities. So if you have a cell with 50% X, 25% O it would look something like

```
XXXXX
XX
 OOOO
```

Also with the current way of doing maybe it would be nicer to omit the empty probability, since at least it would be easy to distinguish empty cells that way.

cognigami (approved):

Issue #21, opened 2022-06-06
Unitary: popping makes quantum object unusable


dstrain115:
Reported by @bartmate 

After popping (measuring) a 1-bit qobject, it behaves differently. First I thought that it will be "set" into the measured value and then I can use it as before. But it seems it is not the case, I cannot use it any more, or at least it will not behave the same way as before. I solved the problem by adding new 1-bit qobjects which are initialized to the measured values.

I think this is uncomfortable. If I cannot use a qobject any more after popping , it would be great, if the library would be able to automatically do this work-around logic.

cognigami:
@bartmate to test on Fox in the Hole.

bartmate:
I deleted the part which always add a new qobject after measurement: see PR #37.

I found two problems:
* The 2nd one is a run-time error. It may or may not be related with the 2nd one.
* The 1st one: It seems there is still some problem with using a 1-bit QObject after a measurement (or maybe my expectation is wrong): The 2nd move is a split which should split the 4th (index=3) hole into 3th and 5th. My expectation is that those 2 holes should have ~.5 - ~.5 probabilites after the split, but it is not the case. The 3th hole has 0 probabilites, while the 5th one has 1.0.  I suspect, this is because the 3rd (index=2) hole was measured beforehand and it fixed the qobject somehow, so later splits cannot modify it.

Please, check.

**1st:**
```
(base) bm@ct:~/01Work/91Unitary/unitary/unitary/examples/fox_in_a_hole$ python3 fox_in_a_hole.py -q -fixseed
Quantum Fox-in-a-hole game.
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2
Where is the fox? (0-4 or q for quit)
2

It seems you have lost :-(. Try again.

History:
State: [0.0, 0.0, 0.0, 0.0, 1.0]
Guess: 2
State: [0.0, 0.0, 0.0, 0.0, 1.0]
Moving (SWAP-based) left from position 4.
State: [0.0, 0.0, 0.0, 1.0, 0.0]
Guess: 2
[...]

PR #28, opened 2022-06-09
Fix post-selection and re-use of popped qubits.

dstrain115:
- This PR fixes post-selection and reuse of qubits.
- Now, after a pop(), the qubit is replaced by a new qubit that is
  initialized to the value of the measurement, and the old qubit
  is moved to an ancilla and post-selected on.
- This PR also includes a force_measurement() function, that can
be used to force a measurement of a particular outcome (for instance,
in replaying a game).

Fixes:#21

### DIFF
--- a/unitary/alpha/quantum_world.py
+++ b/unitary/alpha/quantum_world.py
@@ -125,15 +125,15 @@ class QuantumWorld:
            equal to the count parameter.  Each element will be a list
            of measurement results for each object.
         """
-        if _num_reps is not None:
+        if _num_reps is None:
+            num_reps = self._suggest_num_reps(count)
+        else:
             if _num_reps > 1e6:
                 raise RecursionError(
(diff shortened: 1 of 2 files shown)


cognigami (approved):

unitary/alpha/quantum_world.py
```diff
@@ -97,17 +125,23 @@ def peek(
            equal to the count parameter.  Each element will be a list
            of measurement results for each object.
         """
+        if _num_reps is not None:
```
cantwellc, line 128: nit:
Why the double negative logic? It took me a little bit of extra braining to parse. _num_reps is optional, so will most  likely be None. Seems more straight forward, and easier for me to parse, to be 

if _num_reps is None:
  ...
else
  ...

dstrain115: Done.

PR #29, opened 2022-06-10
Expanding the TicTacToe example to make it a little more like TiqTaqToe

everthemore:
This PR includes some small additions, notably the ability to choose a ruleset (CLASSICAL vs MINIMAL QUANTUM vs FULLY QUANTUM). The default is FULLY QUANTUM, so running this as before ('python ascii_board.py') does not change any of the previous gameplay. The PR also includes the option for arguments (and has a -h with a minimal description of the game + usage).

### DIFF
--- a/unitary/examples/tictactoe/tic_tac_toe.py
+++ b/unitary/examples/tictactoe/tic_tac_toe.py
@@ -167,8 +167,6 @@ class TicTacToe:
             QuditFlip(3, 0, mark.value)(self.squares[move])
             # This square is now no longer empty
             self.empty_squares.discard(move)
-            # This square now is a set by itself
-            self.square_to_set[move] = len(self.square_to_set)
         else:
             # Check if rules allow quantum moves
             if self.rules == TicTacRules.CLASSICAL:
--- a/unitary/examples/tictactoe/tic_tac_toe.py
+++ b/unitary/examples/tictactoe/tic_tac_toe.py
@@ -101,7 +101,7 @@ class TicTacToe:
     or more samples (measurements) of the board.
     """
 
-    def __init__(self, rules: TicTacRules = TicTacRules.FULLY_QUANTUM):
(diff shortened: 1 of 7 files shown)


unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -151,18 +151,54 @@ def move(self, move: str, mark: TicTacSquare) -> TicTacResult:
...
+                            )
+                        return self.result()
+
+            # TicTacSplit first flips the first square before performing a split
+            # If either of the two involved squares is empty, we want to do the
+            # split on that square.
+            if move[0] in self.empty_squares:
+                TicTacSplit(mark)(self.squares[move[0]], self.squares[move[1]])
+            else:
+                TicTacSplit(mark)(self.squares[move[1]], self.squares[move[0]])
```
dstrain115, line 195: Can we add tests for this and the other new functionality?

everthemore:
Apart from updates to docstrings and new tests, this PR now has an important extra change. This should probably have been another PR, but I propose leaving it here instead. Let me know if you'd like me to split this up.

The important change:
- The game now uses the full sqrt-iSWAP gate instead of the specific QuditSplitGate. That means not only are 'XE' and 'EX' swapped, and 'OE' and 'EO', but now also 'XO' and 'OX'. This now matches the existing TiqTaqToe even more closely.

cognigami (approved):

dstrain115:
@everthemore Is full sqrt-iswap really what we want for this game?  It leads to some odd results.  For instance, X splits on 'ab' and then O splits on 'bc' will result in a ~25% chance of X being on 'c', which is unexpected.  I had this originally, but then removed it when I saw a bunch of weird results like that.

dstrain115 (approved):
LGTM, with the caveat that we should add a README.md and the additional rule set in follow-up PRs.

dstrain115 (approved):
LGTM, thanks for adding both options!

PR #33, opened 2022-06-13
Histogram + Probability methods for QWorld

bartmate:
It seems that both examples need some kind of histogram and/or probability functionality, typically for UI purposes.
So I added 3 methods to the class QuantumWorld with which histogram & probabilities can be acquired.


unitary/alpha/quantum_world.py
```diff
@@ -11,13 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Dict, List, Optional, Sequence, Union
```
cognigami, line 14: Out of curiosity, why did you move these lower?

bartmate: I just used pylint, to check the code. For some reason it likes if the standard libraries are imported first. 
So, what I wanted to do is moving "import cirq" lower, but it seems I copy-pasted another line, too :)

unitary/alpha/quantum_world.py
```diff
@@ -11,13 +11,12 @@
...
-import cirq
 import copy
 import enum
+from typing import Dict, List, Optional, Sequence, Union
+import cirq
 
 from unitary.alpha.quantum_object import QuantumObject
-from unitary.alpha.quantum_effect import QuantumEffect
```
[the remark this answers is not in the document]
dstrain115: We probably needed it at some point, then removed the need for it later.

unitary/alpha/quantum_world.py
```diff
@@ -146,3 +145,71 @@ def pop(
...
+            histogram.append({state: 0 for state in range(obj.num_states)})
+        for result in peek_results:
+            for idx in range(len(objects)):
+                histogram[idx][result[idx][0]] += 1
+        return histogram
+
+    def get_probabilities(self,
+                          objects : Optional[Sequence[QuantumObject]] = None,
+                          count: int = 100
+    ) -> List[Dict[int, float]]:
```
dstrain115, line 179: Same here.  The result could be an Enum.

unitary/alpha/quantum_world_test.py
```diff
@@ -151,3 +151,59 @@ def test_pop_not_enough_reps():
...
+    assert histogram[0][1]>10
+    probs = world.get_probabilities()
+    assert len(probs)==1
+    assert len(probs[0])==2
+    assert probs[0][0]>.1
+    assert probs[0][1]>.1
+    bin_probs = world.get_binary_probabilities()
+    assert 0.1 <= bin_probs[0] <= 1.0
+
+    l1 = alpha.QuantumObject("l1", StopLight.YELLOW)
```
dstrain115, line 186: Can we break this up into multiple tests?  It might make it more obvious what you are testing, and they seem independent.

dstrain115 (approved):
LGTM minus a few nits.

unitary/alpha/quantum_world.py
```diff
@@ -146,3 +145,71 @@ def pop(
             self.post_selection[objects[idx]] = post_selection
 
         return results[0]
+
+    def get_histogram(self,
+                      objects: Optional[Sequence[QuantumObject]] = None,
+                      count: int = 100
+    ) -> List[Dict[int, int]]:
```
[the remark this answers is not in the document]
bartmate: In peek I used convert_to_enum=False, so here it should be int.  Later, maybe, we can write a nicer version, which can handle both enum or int, but this always converts the states to int.

unitary/alpha/quantum_world.py
bartmate: In peek I used convert_to_enum=False, so here it should be int.  Later, maybe, we can write a nicer version, which can handle both enum or int, but this always converts the states to int.

unitary/alpha/quantum_world_test.py
bartmate: Done.

unitary/alpha/quantum_world.py
```diff
@@ -146,3 +145,71 @@ def pop(
...
+        for obj_hist in histogram:
+            probabilities.append(
+                {state: obj_hist[state]/count for state in range(len(obj_hist))} )
+        return probabilities
+
+    def get_binary_probabilities(self,
+                                 objects : Optional[Sequence[QuantumObject]] = None,
+                                 count: int = 100
+    ) -> List[float]:
+        """Calculates the probabilities based on measurements (peeks) carried out.
```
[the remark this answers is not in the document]
bartmate: Done.

bartmate:
It seems that some tests in this PR fail, however I do not think it does anything to do with my changes. All tests in quantum_world_test.py passed.

cognigami (approved):
Build failures addressed in PR #36.

bartmate:
Hi.
I think I have no write access to this repo. Could you, please, merge this PR?
Máté

dstrain115:
I added you to write access.  Would you mind updating the branch to verify that tests pass then you can go ahead and merge.

PR #37, opened 2022-06-25
Reusing qobject after fix of #21

bartmate:
DO NOT MERGE YET
See Issue #21 before.


bartmate:
DO NOT MERGE YET
Hm. Maybe I messed up with old libraries... Let me check...

PR #40, opened 2022-07-01
Update castling rules

losos0:


dstrain115 (commented):
Do you think you could explain what the differences are (in the PR description)?  This would help to review it.

losos0:
The existing implementation always measured the target squares (if potentially occupied), while there is supposed to be an EXCLUDED variant that measures and a BASIC variant that doesn't. The BASIC variant is needed because you are allowed to castle without measuring if there is a same-color rook in superposition on the rook's target square.

unitary/quantum_chess/quantum_board.py
```diff
@@ -660,24 +665,72 @@ def _clear_path_ancilla(self, path_qubits, ancilla):
...
+                self.set_castle(sbit, rook_sbit, tbit, rook_tbit)
+                return 1
+
+            self.state = set_nth_bit(sbit, self.state, False)
+            self.state = set_nth_bit(tbit, self.state, True)
+            self.add_entangled(rook_squbit, rook_tqubit)
+            self.circuit.append(qm.normal_move(rook_squbit, rook_tqubit))
+            return 1
+
+        elif variant == enums.MoveVariant.EXCLUDED:
```
dstrain115, line 708: Maybe we could add some comments here explaining the two versions of the castle as you did in the PR?

dstrain115 (approved):

unitary/quantum_chess/quantum_board.py
losos0: Done.

PR #42, opened 2022-07-01
Add ability to combine two QuantumWorlds

dstrain115:
- Combines the QuantumObjects of both worlds,
in addition to the post-selection criteria.

- This functionality is needed for the Quantum RPG.


losos0:
Please move whitespace changes to a separate commit. It's hard to find the relevant diffs

dstrain115:
Whitespace changes merged, PTAL

unitary/alpha/quantum_world.py
```diff
@@ -52,6 +52,16 @@ def add_object(self, obj: QuantumObject):
...
         obj.initial_effect()
 
+    def combine_with(self, other_world: "QuantumWorld"):
+        if self.used_object_keys.intersection(other_world.used_object_keys):
+            raise ValueError("Cannot combine two worlds with overlapping object keys")
+        self.objects.extend(other_world.objects)
+        self.used_object_keys.update(other_world.used_object_keys)
+        self.ancilla_names.update(other_world.ancilla_names)
+        self.post_selection.update(other_world.post_selection)
+        for op in other_world.circuit.all_operations():
```
losos0, line 62: You can write `self.circuit += other.circuit`

unitary/alpha/quantum_world.py
```diff
@@ -52,6 +52,16 @@ def add_object(self, obj: QuantumObject):
...
+            self.circuit.append(op)
+
```
losos0, line 64: Does `self.effect_history` need to be updated? Cleared maybe.

unitary/alpha/quantum_world.py
```diff
@@ -52,6 +52,16 @@ def add_object(self, obj: QuantumObject):
         obj.world = self
         obj.initial_effect()
 
+    def combine_with(self, other_world: "QuantumWorld"):
```
losos0, line 55: What's the status of other_world after the operation? Is it still valid to use?

There should be a comment explaining which object contains the combined world after this is called.

unitary/alpha/quantum_world.py
dstrain115: It's now cleared.

unitary/alpha/quantum_world.py
dstrain115: Added a docstring to explain this.  I also realized the QuantumWorld class is missing a docstring, so added that as well.

I also now explicitly clear the other_world.

unitary/alpha/quantum_world.py
```diff
@@ -66,7 +110,8 @@ def undo_last_effect(self):
         Note that pop() is considered to be an effect for the purposes
         of this call.
         """
-        self.circuit, self.post_selection = self.effect_history.pop()
+        if self.effect_history:
```
losos0, line 113: Throwing an exception seems better if there is no last effect.

losos0 (approved):

unitary/alpha/quantum_world.py
dstrain115: Done.

PR #55, opened 2022-07-07
Add CI checks for mypy and pylint

dstrain115:
- Configuration copied from cirq.

### DIFF
--- a/dev_tools/.pylintrc
+++ b/dev_tools/.pylintrc
@@ -1,0 +1,77 @@
+[MASTER]
+load-plugins=pylint.extensions.docstyle,pylint.extensions.docparams,pylint_copyright_checker
+max-line-length=100
+disable=all
+#ignore-paths=cirq-google/cirq_google/cloud/.*
+ignore-patterns=.*_pb2\.py
+output-format=colorized
+score=no
+reports=no
+enable=
+    anomalous-backslash-in-string,
+    assert-on-tuple,
+    bad-indentation,
+    bad-option-value,
+    bad-reversed-sequence,
+    bad-super-call,
+    consider-merging-isinstance,
+    continue-in-finally,
+    dangerous-default-value,
+    docstyle,
+    duplicate-argument-name,
+    expression-not-assigned,
+    f-string-without-interpolation
+    function-redefined,
+    inconsistent-mro,
+    init-is-generator,
+    line-too-long,
+    lost-exception,
+    missing-kwoa,
+    missing-param-doc,
+    missing-raises-doc,
+    mixed-indentation,
+    mixed-line-endings,
+    no-value-for-parameter,
+    nonexistent-operator,
+    not-in-loop,
+    pointless-statement,
+    redefined-builtin,
+    relative-import,
+    return-arg-in-generator,
+    return-in-init,
+    return-outside-function,
+    simplifiable-if-statement,
+    singleton-comparison,
+    syntax-error,
+    too-many-function-args,
+    trailing-whitespace,
+    undefined-variable,
+    unexpected-keyword-arg,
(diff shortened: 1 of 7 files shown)


losos0:
FWIW pytype has a lot less problems than mypy on the quantum chess codebase.

dstrain115:
> FWIW pytype has a lot less problems than mypy on the quantum chess codebase.

Cirq and other codebases we use all use mypy, so I would like to stay consistent.

dev_tools/mypy.ini
```diff
@@ -0,0 +1,27 @@
+[mypy]
+show_error_codes = true
+
+[mypy-__main__]
+follow_imports = silent
+ignore_missing_imports = true
+
+# 3rd-party libs for which we don't have stubs
+[mypy-apiclient.*,freezegun.*,matplotlib.*,mpl_toolkits,multiprocessing.dummy,oauth2client.*,pandas.*,proto.*,pytest.*,scipy.*,sortedcontainers.*,setuptools.*,pylatex.*,networkx.*,qiskit.*,pypandoc.*,ply.*,_pytest.*,google.api.*,google.api_core.*,grpc.*,google.auth.*,google.oauth2.*,google.protobuf.text_format.*,quimb.*,pyquil.*,google.cloud.*,filelock.*,codeowners.*,tqdm.*,importlib_metadata.*,google.colab.*,IPython.*,astroid.*,pylint.*,cirq.*,flask.*,numpy.*,cirq_google.*]
```
losos0, line 9: Wow, do you really have to maintain a list of transitive dependencies to use mypy? I think this will be a problem (if the check is enforced). The set of dependencies can change at any time since the package manager generally downloads the latest versions modulo constraints. So this will lead to CI errors on pull requests unrelated to the actual changes whenever a new transitive dependency appears.

dstrain115: It's not transitive dependencies, it's only direct dependencies.  CI errors for this will only be if a new dependency appears that was introduced in the actual PR.

Hopefully I can figure out how to get stubs for some of the libraries as well.
Let me see if I can clean this up.

losos0: What do you mean it's not transitive dependencies? For example `freezegun`, `protobuf`, or `flask` are not referenced by the Unitary codebase.

dstrain115: Those were all from cirq, where I copied the continuous integration test from.  So, there were a lot of extra unneeded dependencies in there.  If you refresh to the latest commit, I have pruned it way down (there may still be a few extra unneeded ones, but I got rid of a bunch of them).

requirements.txt
```diff
@@ -4,5 +4,12 @@ cirq-google>=0.14.0
...
 seaborn
 sphinx
+
+# dev tools
 ipython
 black
+mypy
+pylint
+pytest
+pytest-cov
```
losos0, line 14: Why install coverage tools?

dev_tools/.pylintrc
```diff
@@ -0,0 +1,73 @@
+[MASTER]
+load-plugins=pylint.extensions.docstyle,pylint.extensions.docparams,pylint_copyright_checker
+max-line-length=100
```
losos0, line 3: 88 to agree with black?

dev_tools/mypy.ini
```diff
@@ -0,0 +1,16 @@
+[mypy]
+show_error_codes = true
+
+[mypy-__main__]
```
losos0, line 4: What does this mean?

cantwellc (commented):
I'm not sure how useful I'll be here. I can work my way through python, but I'm no super versed in all of the tooling.

PR #60, opened 2022-07-20
First batch of type issues from mypy

dstrain115:
- cirq.Qid does not have a name, so changing to NamedQubit in typing.
- Also misc typing fixes all over.
- Remove gateset in engine.sampler, since it no longer exists in Cirq 1.0.

### DIFF
--- a/unitary/engine_utils.py
+++ b/unitary/engine_utils.py
@@ -18,7 +18,7 @@ import math
 import os
 import uuid
 from dataclasses import dataclass, field
-from typing import List, Any, Optional, Callable, Dict, Union
+from typing import List, Any, Optional, Callable, Dict, Union, Sequence
 
 import numpy as np
 
--- a/unitary/engine_utils.py
+++ b/unitary/engine_utils.py
@@ -164,7 +164,7 @@ class ZerosSampler(work.Sampler):
             program: 'cirq.Circuit',
             params: 'cirq.Sweepable',
             repetitions: int = 1,
-    ) -> List['cirq.Result']:
+    ) -> Sequence['cirq.Result']:
         assert isinstance(program, circuits.Circuit)
         meas = list(program.findall_operations_with_gate_type(
             ops.MeasurementGate))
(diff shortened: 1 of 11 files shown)


unitary/engine_utils.py
```diff
@@ -247,13 +247,7 @@ def engine(self):
...
 
     def get_sampler(self, gateset: str = None):
-        if gateset == 'sycamore':
-            gateset = cg.SYC_GATESET
-        elif gateset == 'sqrt-iswap':
-            gateset = cg.SQRT_ISWAP_GATESET
-        else:
-            raise ValueError("Unknown gateset {}".format(gateset))
-        return self.engine.sampler(processor_id=self.processor_id, gate_set=gateset)
+        return self.engine.sampler(processor_id=self.processor_id)
```
losos0, line 250: What's this part? Doesn't seem related to type checking.

unitary/quantum_chess/quantum_board.py
```diff
@@ -135,7 +135,9 @@ def with_state(self, basis_state: int, reset_move_history=True) -> "CirqBoard":
         self.post_selection = {}
         self.circuit = cirq.Circuit()
         self.ancilla_count = 0
-        initial_square_probs = (float(nth_bit_of(i, basis_state)) for i in range(64))
+        initial_square_probs = list(
```
losos0, line 138: Use a list comprehension

unitary/engine_utils.py
dstrain115: gateset parameter no longer exists in Cirq 1.0.  Added to the PR description.

I can split it out into a different PR if you want.

unitary/quantum_chess/quantum_board.py
dstrain115: Done.

unitary/engine_utils.py
losos0: That's all right. A different commit would have been nice though.

unitary/quantum_chess/move.py
```diff
@@ -185,9 +185,9 @@ def to_string(self, include_type=False) -> str:
           include_type: also include the move type/variant in the string
         """
         movestr = self.source + self.target
-        if self.is_split_move():
+        if self.is_split_move() and self.target2:
```
losos0, line 188: This change seems suboptimal since if you had a wrongly constructed split move, then the evidence would be covered up when printing it as a string. I think you could do `str(self.target2)` instead

losos0 (approved):

unitary/quantum_chess/move.py
dstrain115: Good idea.  Done.

PR #64, opened 2022-08-13
[Chess] Improve behavior on some edge cases

losos0:
Squares measured as 0% or 100% occupied may actually be in a superposition due to noisy measurements (see commit message for more detail).


dstrain115 (approved):
LGTM

dstrain115 (commented):
It would be nice to add some tests for these edge cases though.

losos0:
Added tests.

PR #66, opened 2022-08-24
Classically keep track of non-empty holes for the fox in the hole example

everthemore:
As the game progresses, it becomes increasingly slow (up to several seconds for games of 8 moves) to play. Partly, this is because of the game re-computing the probability distribution for the state at every move in order to determine a 'source' from which the fox can move. 

This PR will circumvent that by keeping track of the empty holes classically.

**Important**: The real issue is an underlying one in the way this sampling happens. After pop()'s, ancillary qubits are added to the quantum world, which are subsequently also being sampled (I think?). That may not be the desired behaviour.

### DIFF
--- a/unitary/examples/fox_in_a_hole/fox_in_a_hole.py
+++ b/unitary/examples/fox_in_a_hole/fox_in_a_hole.py
@@ -174,6 +174,10 @@ class QuantumGame(Game):
             holes.append(hole)
         self.state = (QuantumWorld(holes), holes)
 
+        # Track empty holes
+        self.empty_holes = np.ones(self.hole_nr)
+        self.empty_holes[index] = 0
+
     def state_to_string(self):
         return str(self.state[0].get_binary_probabilities(objects=self.state[1]))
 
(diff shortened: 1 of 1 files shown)


everthemore:
Here is an example of the behaviour, and a simple timing profile (over the entire length of the game)

Screenshot 2022-08-24 at 12 54 40

Screenshot 2022-08-24 at 12 53 55

bartmate:
Hi. 
In my understanding, it is planned that a similar kind of optimization will be implemented in the Unitary library itself. If this is the case I would not add this change to the game, as it makes it more complex, although it was planned as a simple example for using the library. Also, it uses some kind of "meta"-knowledge about the underlying quantum operations (i.e. after a q-move which qbit will/will not have guaranteedly zero coeff/probability). Maybe, this is trivial for us, but a game-developer user of the library without quantum knowledge may not know such details. 
If any of you does not agree, let's have a discussion about this on the next Tuesday meeting. I am open to change my mind.
Máté

everthemore:
You raise two important points (doing this in unitary & this making use of quantum knowledge this way). As far as I am concerned this should probably settle the issue in favor of *not* merging this. It was an attempt at speeding this up temporarily, though as I noted the real issue is a deeper one. Let's put our resources on fixing that instead.

PR #68, opened 2022-09-08
Sparse state vector simulator

losos0:
Add a sparse state vector simulator which can be used as the sampler for a QuantumWorld. This speeds up simulation for circuits which use only a tiny fraction of the 2^n possible states. In addition to calculating on only the used states instead of all states, a post-selection operator is inserted into circuits when performing a measurement using the sparse simulator, which prunes the states not conforming with the measurement outcome.

Not sure what kind of unit tests we might want for sparse_vector_simulator. Is there like a standard test suite for Cirq simulators?

### DIFF
--- a/unitary/alpha/quantum_world.py
+++ b/unitary/alpha/quantum_world.py
@@ -17,6 +17,7 @@ from typing import Dict, List, Optional, Sequence, Union
 import cirq
 
 from unitary.alpha.quantum_object import QuantumObject
+from unitary.alpha.sparse_vector_simulator import PostSelectOperation, SparseSimulator
 
 
 class QuantumWorld:
@@ -42,6 +43,7 @@ class QuantumWorld:
 
         self.clear()
         self.sampler = sampler
+        self.use_sparse = isinstance(sampler, SparseSimulator)
 
         if isinstance(objects, QuantumObject):
             objects = [objects]
@@ -86,6 +88,8 @@ class QuantumWorld:
         """
         if self.used_object_keys.intersection(other_world.used_object_keys):
             raise ValueError("Cannot combine two worlds with overlapping object keys")
+        if self.use_sparse != other_world.use_sparse:
+            raise ValueError("Cannot combine sparse simulator world with non-sparse")
         self.objects.extend(other_world.objects)
         self.used_object_keys.update(other_world.used_object_keys)
         self.ancilla_names.update(other_world.ancilla_names)
@@ -96,13 +100,29 @@ class QuantumWorld:
         # Clear the other world so that objects cannot be used from that world.
         other_world.clear()
 
+    def _append_op(self, op: cirq.Operation):
+        """Add the operation in a way designed to speed execution.
+
+        For the sparse simulator post-selections should be as early as possible to cut
+        down the state size. Also X's since they don't increase the size.
+        """
+        if not self.use_sparse:
+            self.circuit.append(op)
+            return
+
+        if isinstance(op, PostSelectOperation) or op.gate is cirq.X:
+            strategy = cirq.InsertStrategy.EARLIEST
+        else:
+            strategy = cirq.InsertStrategy.NEW
+        self.circuit.append(op, strategy=strategy)
+
     def add_effect(self, op_list: List[cirq.Operation]):
         """Adds an operation to the current circuit."""
         self.effect_history.append(
             (self.circuit.copy(), copy.copy(self.post_selection))
         )
         for op in op_list:
-            self.circuit.append(op)
+            self._append_op(op)
 
     def undo_last_effect(self):
         """Restores the `QuantumWorld` to the state before the last effect.
@@ -124,6 +144,8 @@ class QuantumWorld:
         Noise and error mitigation will discard reps, so increase the total
         number of repetitions to compensate.
         """
+        if self.use_sparse:
+            return sample_size
         if len(self.post_selection) >= 1:
             sample_size <<= len(self.post_selection) + 1
         if sample_size < 100:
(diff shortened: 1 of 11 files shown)


unitary/alpha/sparse_vector_simulator.py
```diff
@@ -0,0 +1,121 @@
...
+        self._states = np.array([0], dtype=object)
+        self._amplitudes = np.array([1], dtype=np.complex128)
+
+    def copy(self):
+        raise NotImplementedError
+
+    def _act_on_fallback_(self, action, qubits, allow_decompose):
+        if action.gate is cirq.X:
+            self._states ^= 1 << self.qubit_map[qubits[0]]
+        else:
```
dstrain115, line 34: Yikes.  This else block is pretty dense.  Some line comments could help here.

unitary/alpha/sparse_vector_simulator.py
```diff
@@ -0,0 +1,121 @@
...
+    ):
+        assert initial_state == 0
+        return SparseSimulationState(qubits=qubits)
+
+    # abstract method of SimulatorBase
+    def _create_step_result(self, sim_state):
+        return SparseSimulatorStep(sim_state=sim_state)
+
+
+class PostSelectOperation(cirq.Operation):
```
dstrain115, line 102: This needs documentation.  Also, I would move this out into its own file.

unitary/alpha/sparse_vector_simulator.py
```diff
@@ -0,0 +1,121 @@
...
+        return out
+
+    def post_select(self, qubit, value):
+        assert value in (0, 1)
+        mask = 1 << self.qubit_map[qubit]
+        value *= mask
+        nonzero_indices = np.nonzero(self._states & mask == value)
+        self._states = self._states[nonzero_indices]
+        self._amplitudes = self._amplitudes[nonzero_indices]
+        self._amplitudes /= np.linalg.norm(self._amplitudes)
```
dstrain115, line 77: What happens if a value is post_selected that is invalid (ie. the norm is zero)?  Maybe we should raise an error here?

unitary/alpha/sparse_vector_simulator.py
```diff
@@ -0,0 +1,121 @@
...
+    # abstract method of SimulatorBase
+    def _create_partial_simulation_state(
+        self, initial_state, qubits, logs=None, classical_data=None
+    ):
+        assert initial_state == 0
+        return SparseSimulationState(qubits=qubits)
+
+    # abstract method of SimulatorBase
+    def _create_step_result(self, sim_state):
+        return SparseSimulatorStep(sim_state=sim_state)
```
dstrain115, line 99: I would definitely add some unit tests.  You can see the unit tests for sparse simulator here:
https://github.com/quantumlib/Cirq/blob/master/cirq-core/cirq/sim/sparse_simulator_test.py

I suspect we do not need such extensive tests, but we should test all supported circuits.

You could also test something like this versus the cirq Simulator?
https://quantumai.google/reference/python/cirq/testing/random_circuit

unitary/alpha/sparse_vector_simulator.py
```diff
@@ -0,0 +1,121 @@
+"""Noiseless simulator using a sparse state vector.
+
+Just enough features to support Unitary are implemented.
```
dstrain115, line 3: I think I would clarify this statement and say what features are supported.  It looks like this supports all cirq gates that have unitary actions as well as the post selection operator.  Correct?

dstrain115 (commented):
I would change the PR description to explain what was added and the features this PR now supports.

unitary/alpha/sparse_vector_simulator.py
```diff
@@ -0,0 +1,121 @@
...
+
+Does not work with 3+-state qudits. TODO: integrate with qutrit-to-qubit conversion
+when that is available.
+"""
+import cirq
+import numpy as np
+
+
+# TODO: Is this a good number?
```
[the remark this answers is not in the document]
losos0: Done.

unitary/alpha/sparse_vector_simulator.py
losos0: Correct.

unitary/alpha/quantum_world.py
```diff
@@ -96,13 +100,29 @@ def combine_with(self, other_world: "QuantumWorld"):
...
+
+        For the sparse simulator post-selections should be as early as possible to cut
+        down the state size. Also X's since they don't increase the size.
+        """
+        if not self.use_sparse:
+            self.circuit.append(op)
+            return
+
+        if isinstance(op, PostSelectOperation) or op.gate is cirq.X:
+            strategy = cirq.InsertStrategy.EARLIEST
```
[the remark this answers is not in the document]
losos0: If everyone is EARLIEST, no one is... I add most operations with NEW to have them added at the end, so that they don't get placed ahead of post-selection operations which work better if they are earlier.

unitary/alpha/sparse_vector_simulator.py
losos0: Added documentation. Why a separate file? All the classes in this file are closely coupled to each other.

unitary/alpha/sparse_vector_simulator.py
losos0: Added an exception InvalidPostSelectionError.

unitary/alpha/sparse_vector_simulator.py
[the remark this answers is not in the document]
dstrain115: You can probably get rid of the TODO.

dstrain115 (approved):
LGTM.  Sorry about the delay in re-review.

dstrain115:
@smitsanghavi Do you want to take a look at this since you have worked on cirq classical simulation?

bartmate (approved):

PR #70, opened 2022-11-03
TicTacToe on qubits

smitsanghavi:
Dirty hack


mhucka:
@smitsanghavi Thank you for this draft contribution. It has been quite some time since there was any activity on this. Could you let us know if you are interested and able to finish this and turn it into a full pull request?

mhucka:
> If there is a need to generalize this to worlds where each qudit has a different dimension, https://github.com/quantumlib/unitary/issues/77 is the tracking issue for that. I can work on it after a couple of weeks if it seems useful.

@dstrain115 You would be a better person to answer that question, as you are the leading quantum gamer, I believe :-).

PR #71, opened 2022-11-30
Initial Classes for Quantum RPG

dstrain115:
- This PR includes the initial Qaracter class (ie. Q-Character) that defines how a character's health points operate.
- Each Qaracter has one health point (HP) per qubit/QuantumObject.
- Each HP is initially unmeasured, and, once measured, determines whether that point is Healthy (1) or Hurt (0).

### DIFF
--- a/unitary/examples/quantum_rpg/qaracter.py
+++ b/unitary/examples/quantum_rpg/qaracter.py
@@ -1,0 +1,140 @@
+from typing import Callable, Dict, List, Optional, Union
+import random
+from unitary import alpha
+from unitary.examples.quantum_rpg import enums
+
+
+class Qaracter(alpha.QuantumWorld):
+    """Base class for quantum RPG characters.
+
+  Quantum characters, or `Qaracter`s are made of
+  `QuantumObject`s.  The level of a quantum character
+  is equivalent to the number of `QuantumObject`s
+  (not including any ancillae added for measurement post-selection).
+
+  Qaracters have one hit point (HP) per level.  Qaracters are
+  considered DOWN if they have over half their qubits measured
+  as zero.  Qaracters are considered ESCAPED if they have all
+  their qubits measured and at least half are measured as one.
+  If neither of these is true, the Qaracter is considered to
+  be ACTIVE.
+
+  Additional HP can be added with the add_hp() function.
+  Additional quantum effects can be added with the add_quantum_effects()
+  function.
+
+  Hit point values can be sampled (a random result pulled from
+  the simulated distribution without measuring, which is possible
+  only in simulation) with sample(hp_num, False).
+
+  Hit point values can be measured by calling sample(hp_num, True).
+
+
+  Args:
+      name: name of this character as a string.
+  """
+
+    def __init__(self, name: str):
+        super().__init__()
+        self.name = name
+        self.health_status: Dict[alpha.QuantumObject, int] = {}
+        self.level = 0
+        self.add_hp()
+
+    def is_npc(self) -> bool:
+        """Returns Trus if a non-player or False if a player.
+
+    Inheritors of NPCs should override this function.
+    """
+        return False
+
+    def quantum_object_name(self, hp_num: int) -> str:
+        """Canonical name of a QuantumObject for this Qaracter.
+
+      Represents the n-th HP of a Qaracter.
+
+      Args:
+          hp_num: Number of the health point requested.  Health points
+              start at 1 and go up to the Qaracter's level (inclusive).
+      """
+        return f'{self.name}_{hp_num}'
+
+    def add_hp(self) -> alpha.QuantumObject:
+        """Adds an additional Health Point (HP to the Qaracter.
+
+      Each HP added to the Qaracter is an additional QuantumObject
+      with no effects (i.e. is a qubit with no operations).
+      Adding a HP also increases the Qaracter's level.
+      """
+        self.level += 1
+        obj = alpha.QuantumObject(self.quantum_object_name(self.level),
+                                  enums.HealthPoint.HURT)
+        self.add_object(obj)
+        return obj
+
+    def get_hp(self, hp_name: str) -> alpha.QuantumObject:
+        """Retrieves a QuantumObject with the specified name."""
+        for obj in self.objects:
+            if hp_name == obj.name:
+                return obj
+
+    def active_qubits(self) -> List[str]:
+        """Returns the names of all active (non-measured) quantum objects (HP)."""
+        return [
+            self.quantum_object_name(i)
+            for i in range(1, self.level + 1)
+            if self.quantum_object_name(i) not in self.health_status
+        ]
+
+    def add_quantum_effect(self, effect, quantum_name: Union[int, str]):
+        """Adds a Quantum Effect to a specific quantum object."""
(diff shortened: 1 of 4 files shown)


unitary/examples/quantum_rpg/qaracter.py
```diff
@@ -0,0 +1,140 @@
...
+      Represents the n-th HP of a Qaracter.
+
+      Args:
+          hp_num: Number of the health point requested.  Health points
+              start at 1 and go up to the Qaracter's level (inclusive).
+      """
+        return f'{self.name}_{hp_num}'
+
+    def add_hp(self) -> alpha.QuantumObject:
+        """Adds an additional Health Point (HP to the Qaracter.
```
smitsanghavi, line 63:
```suggestion
        """Adds an additional Health Point (HP) to the Qaracter.
```

unitary/examples/quantum_rpg/qaracter.py
```diff
@@ -0,0 +1,140 @@
...
+
+      Each HP added to the Qaracter is an additional QuantumObject
+      with no effects (i.e. is a qubit with no operations).
+      Adding a HP also increases the Qaracter's level.
+      """
+        self.level += 1
+        obj = alpha.QuantumObject(self.quantum_object_name(self.level),
+                                  enums.HealthPoint.HURT)
```
smitsanghavi, line 71: I don't know anything about the intended gameplay, but wouldn't HEALTHY be a more intuitive default for a newly added HP?

unitary/examples/quantum_rpg/qaracter.py
```diff
@@ -0,0 +1,140 @@
...
+        """Adds a Quantum Effect to a specific quantum object."""
+        if isinstance(quantum_name, int):
+            quantum_name = self.quantum_object_name(quantum_name)
+        hp_obj = self.get_hp(quantum_name)
+        effect(hp_obj)
+
+    def is_down(self) -> bool:
+        """Returns True if over half of the HP have been measured as `HURT`."""
+        damage = len(
+            [q for q in self.health_status if self.health_status[q] == 0])
```
smitsanghavi, line 99: `sum(status == 0 for status in self.health_status.values())` would be cleaner (here and for `is_escaped`)

unitary/examples/quantum_rpg/qaracter.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def is_escaped(self) -> bool:
+        """Returns True if all HP are measured and at least half are `HEALTHY`."""
+        virtue = len(
+            [q for q in self.health_status if self.health_status[q] == 1])
+        return len(
+            self.health_status) == self.level and virtue > (self.level / 2)
+
+    def is_inactive(self) -> bool:
+        """Returns True if all HP are measured."""
```
smitsanghavi, line 110: This docstring does not necessarily hold true for the is_down state.

unitary/examples/quantum_rpg/qaracter.py
```diff
@@ -0,0 +1,140 @@
...
+        return self.is_down() or self.is_escaped()
+
+    def status_line(self) -> str:
+        """Returns a one-line string summarizing the Qaracter's HP status."""
+        damage = len(
```
smitsanghavi, line 115: would be nice to add `get_damage` and `get_virtue` methods to avoid duplication of this logic

unitary/examples/quantum_rpg/qaracter.py
```diff
@@ -0,0 +1,140 @@
...
+            [q for q in self.health_status if self.health_status[q] == 0])
+        virtue = len(
+            [q for q in self.health_status if self.health_status[q] == 1])
+        unknown = self.level - damage - virtue
+        down = ' *DOWN* ' if self.is_down() else ''
+        return f'{self.level}QP ({virtue}|1> {damage}|0> {unknown}?){down}'
```
smitsanghavi, line 121: Should we also print `ESCAPED` when `is_escaped` is True?

smitsanghavi (changes requested):

unitary/examples/quantum_rpg/qaracter.py
dstrain115: Possibly, but I think it makes more sense from a quantum perspective to start at 0.  The 'qaracter's character sheet corresponds to quantum circuit that acts on a 'blank sheet' of all 0 HPs.

unitary/examples/quantum_rpg/qaracter.py
dstrain115: Good call, done.

unitary/examples/quantum_rpg/qaracter.py
dstrain115: Changed.

unitary/examples/quantum_rpg/qaracter.py
dstrain115: Good idea.  Added damage and virtue as properties.

unitary/examples/quantum_rpg/qaracter.py
dstrain115: Done.

dstrain115 (commented):
Comments addressed.  Thanks for the quick review!

unitary/examples/quantum_rpg/qaracter.py
```diff
@@ -93,44 +93,49 @@ def add_quantum_effect(self, effect, quantum_name: Union[int, str]):
...
+
     def is_down(self) -> bool:
         """Returns True if over half of the HP have been measured as `HURT`."""
-        damage = len(
-            [q for q in self.health_status if self.health_status[q] == 0])
-        return damage > (self.level / 2)
+        return self.damage > (self.level / 2)
 
     def is_escaped(self) -> bool:
         """Returns True if all HP are measured and at least half are `HEALTHY`."""
```
smitsanghavi, line 111: Neither is_escaped or is_down is True right now for `damage == virtue == level/2`. Maybe add a test for an even number of HP.

unitary/examples/quantum_rpg/qaracter.py
```diff
@@ -93,44 +93,49 @@ def add_quantum_effect(self, effect, quantum_name: Union[int, str]):
...
-        virtue = len(
-            [q for q in self.health_status if self.health_status[q] == 1])
         return len(
-            self.health_status) == self.level and virtue > (self.level / 2)
+            self.health_status) == self.level and self.virtue > (self.level / 2)
 
     def is_inactive(self) -> bool:
```
smitsanghavi, line 115: optional nit. Would it be more readable to have an `is_active` method and `not` it as needed rather than possibly having to do `not is_inactive` for checking if a qaracter is active.

smitsanghavi (changes requested):

smitsanghavi (approved):

unitary/examples/quantum_rpg/qaracter.py
dstrain115: Fixed and added case for even qubits.  Thanks for catching it!

unitary/examples/quantum_rpg/qaracter.py
dstrain115: Done.  Remove double negative and changed to is_active().

PR #73, opened 2022-12-05
Update Qaracter.is_active docstring after the latest change

smitsanghavi:
The docstring still explains the `is_inactive` method.


dstrain115 (approved):
Thanks for following up and fixing my carelessness!

PR #74, opened 2022-12-05
Add two example player classes to Quantum RPG

dstrain115:
- This PR adds two very simple player classes to the quantum RPG.
- This PR is advance of a feature that can demonstrate an example battle.

### DIFF
--- a/unitary/examples/quantum_rpg/classes.py
+++ b/unitary/examples/quantum_rpg/classes.py
@@ -1,0 +1,36 @@
+from typing import Any, Callable, Dict
+
+from unitary import alpha
+from unitary.examples.quantum_rpg import qaracter
+
+
+class Engineer(qaracter.Qaracter):
+    """Example player class that can do an X gate.
+
+    Engineeers can do an X gate at level 1 and a H gate at level 3.
+    """
+
+    def actions(self) -> Dict[str, Callable]:
+        actions = {
+            'x': lambda monster, qubit: monster.add_effect(alpha.Flip(), qubit)
+        }
+        if self.level >= 3:
+            actions['h'] = lambda monster, qubit: monster.add_effect(
+                alpha.Superposition(), qubit)
+        return actions
+
+
+class Analyst(qaracter.Qaracter):
+    """Example player class that can sample and measure.
+
+    Analysts can sample ('s') or measure ('m').
+    """
+
+    def actions(self) -> Dict[str, Callable]:
+        return {
+            's':
+                lambda monster, qubit: print(
+                    f'Sample result {monster.sample(qubit, False)}'),
+            'm':
+                lambda monster, qubit: monster.sample(qubit, True)
+        }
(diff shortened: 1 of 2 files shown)


unitary/examples/quantum_rpg/classes_test.py
```diff
@@ -0,0 +1,18 @@
+import unitary.examples.quantum_rpg.classes as classes
```
smitsanghavi, line 1: Can we have some tests for the effects of the actions?

unitary/examples/quantum_rpg/classes.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any, Callable, Dict
+
+from unitary import alpha
+from unitary.examples.quantum_rpg import qaracter
+
+
+class Engineer(qaracter.Qaracter):
+    """Example player class that can do an X gate.
+
+    Engineeers can do an X gate at level 1 and a H gate at level 3.
```
smitsanghavi, line 10: Just curious, how is an H gate more powerful than X in a battle?

smitsanghavi (changes requested):

unitary/examples/quantum_rpg/classes.py
dstrain115: I am not sure it is more powerful.  This is just an example, but I was thinking that the first few battles would essentially be 'classical', and the Hadamard allows you to handle enemies that are in the Clifford space.

unitary/examples/quantum_rpg/classes_test.py
dstrain115: Done.  Good call.  This caught two issues.

unitary/examples/quantum_rpg/classes.py
```diff
@@ -29,8 +31,9 @@ class Analyst(qaracter.Qaracter):
     def actions(self) -> Dict[str, Callable]:
         return {
             's':
-                lambda monster, qubit: print(
-                    f'Sample result {monster.sample(qubit, False)}'),
+                lambda monster, qubit:
+                f'Sample result {monster.sample(monster.quantum_object_name(qubit), False)}',
```
smitsanghavi, line 35: Why remove the `print`?

smitsanghavi (changes requested):

unitary/examples/quantum_rpg/classes.py
dstrain115: Primarily to make it testable.  That way it returns the string and the caller can do what it wants with it.  I suppose in the future we could also display it in a different way (like a javascript based window or something), but I would imagine that's pretty far off.

smitsanghavi: Somewhat nitpicky but if the caller is expected to do what they want with the sample output, maybe the string message should also be controlled by the caller. `s` can just return the sampled value similar to `m`.

smitsanghavi (approved):

unitary/examples/quantum_rpg/classes.py
dstrain115: I am not sure this is a good idea.  For the example of a sample vs measure, the caller needs to understand the difference between them then, since they are both returning a measurement result, so the caller needs to understand the semantic difference between 'm' and 's' which should really be the logic within the class.

I am going to skip this comment for now.  I have a suspicion we will need to revisit this logic anyway once things progress.

PR #75, opened 2022-12-16
Quantum RPG Battle Engine

dstrain115:
- Add class to actually run a Quantum RPG battle.
- This pits players versus NPCs in a battle.
- Player actions and NPC actions can be triggered via functions.
- Also adds a simple test NPC.

At this point, the MVP of the RPG is almost playable, though some work is needed to initialize and run the battle (plus make it more interesting).

### DIFF
--- a/unitary/examples/quantum_rpg/battle.py
+++ b/unitary/examples/quantum_rpg/battle.py
@@ -63,17 +63,6 @@ class Battle:
             print(status, file=self.file)
         print('-----------------------------------------------', file=self.file)
 
-    def _get_user_input(self, prompt: str, user_input: Optional[List[str]],
-                        idx: int):
-        """Helper function to retrieve user input.
-
-        If an array of strings is given, retrieve the next input
-        from the array.  Otherwise, prompt the user using raw input.
-        """
-        if user_input and idx < len(user_input):
-            return user_input[idx]
-        return input(prompt)
-
     def take_player_turn(self, user_input: Optional[List[str]] = None):
         """Take a player's turn and record results in the battle.
 
@@ -86,7 +75,14 @@ class Battle:
             user_input: List of strings that substitute for the user's
                 raw input.
         """
-        idx = 0
+
+        # If user input is provided as an argument, then use that.
+        # Otherwise, prompt from raw input.
+        if user_input is not None:
+            user_input = iter(user_input)
+            get_user_input = lambda _: next(user_input)
+        else:
+            get_user_input = input
 
         for current_player in self.player_side:
             self.print_screen()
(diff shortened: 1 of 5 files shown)


unitary/examples/quantum_rpg/npcs.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    def is_npc(self):
+        return True
+
+
+class Observer(Npc):
+    """Simple test NPC that measures a random qubit each turn."""
+
+    def is_npc(self):
```
losos0, line 19: This override is not needed

unitary/examples/quantum_rpg/battle.py
```diff
@@ -0,0 +1,132 @@
...
+        1) Retrieve the possible actions from the player.
+        2) Prompt the player for which action to use.
+        3) Prompt the player which NPC and qubit to target.
+        4) Call the player's action to perform the action.
+
+        Args:
+            user_input: List of strings that substitute for the user's
+                raw input.
+        """
+        idx = 0
```
losos0, line 89: How about
```
if user_input is not None:
   user_input = iter(user_input)
   get_user_input = lambda unused_prompt: next(user_input)
else:
   get_user_input = input
```

unitary/examples/quantum_rpg/npcs.py
dstrain115: Removed.

unitary/examples/quantum_rpg/battle.py
dstrain115: Great idea, very elegant.  Done.

unitary/examples/quantum_rpg/battle.py
```diff
@@ -0,0 +1,120 @@
...
+            print(f'{current_player.name} turn:', file=self.file)
+            if not current_player.is_active():
+                print(f'{current_player.name} is DOWN!', file=self.file)
+                continue
+            actions = current_player.actions()
+            for key in actions:
+                print(key, file=self.file)
+            action = get_user_input('Choose your action: ')
+            if action in current_player.actions():
+                monster = int( get_user_input('Which enemy number: ')) + 1
```
smitsanghavi, line 98: Shouldn't this be subtracting 1 from the input?

unitary/examples/quantum_rpg/battle.py
```diff
@@ -0,0 +1,120 @@
...
+                if monster < len(self.enemy_side):
```
smitsanghavi, line 99: Print an error when an invalid monster number is provided

smitsanghavi (changes requested):

unitary/examples/quantum_rpg/battle.py
dstrain115: Good catch!  Fixed and added more unit tests for this case.

unitary/examples/quantum_rpg/battle.py
dstrain115: Done (and tested more thoroughly)

dstrain115:
Thanks for the reviews everyone.  Probably last update before 2023, but hope to push on this some more next year!

smitsanghavi (approved):

smitsanghavi:
Looks awesome, excited to explore what all can be done with this framing!

PR #80, opened 2023-02-14
Add effect_fraction and Phase effect

dstrain115:
- This PR adds an optional parameter to the Flip effect that allows a user to have a partial Flip.  This in effect is a X**t gate.
- This PR also adds a Phase effect, which is a Z**t gate (with default t=1).

### DIFF
--- a/unitary/alpha/qubit_effects.py
+++ b/unitary/alpha/qubit_effects.py
@@ -62,7 +62,7 @@ class Phase(QuantumEffect):
     on the |0> and |1> states, but can change the phase (complex sign)
     of states in superpositions.
 
-    A full phase is the same as a rotation around the Z axis, or a Z gate.
+    A full phase is the same as a 180° rotation around the Z axis, or a Z gate.
     For a partial phase, use the `effect_fraction` argument.
 
     These effects will be cumulative.  For instance, two quarter
--- a/unitary/alpha/qubit_effects.py
+++ b/unitary/alpha/qubit_effects.py
@@ -22,14 +22,71 @@ from unitary.alpha.quantum_effect import QuantumEffect
 
 
 class Flip(QuantumEffect):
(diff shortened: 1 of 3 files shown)


unitary/alpha/qubit_effects_test.py
```diff
@@ -30,6 +30,33 @@ def test_flip(simulator, compile_to_qubits):
...
 
+@pytest.mark.parametrize("compile_to_qubits", [False, True])
+@pytest.mark.parametrize("simulator", [cirq.Simulator, SparseSimulator])
+def test_partial_flip(simulator, compile_to_qubits):
+    board = alpha.QuantumWorld(sampler=simulator(), compile_to_qubits=compile_to_qubits)
+    piece = alpha.QuantumObject("t", 0)
+    board.add_object(piece)
+    alpha.Flip(effect_fraction=0.25)(piece)
+    assert board.circuit == cirq.Circuit(cirq.X(cirq.NamedQubit("t")) ** 0.25)
+
```
smitsanghavi, line 41: Add a test to verify the cumulative effect of multiple fractional ops is as expected?

unitary/alpha/qubit_effects.py
```diff
@@ -22,14 +22,71 @@
...
+
+class Phase(QuantumEffect):
+    """Phases a qubit from |+> to |-> and vice versa.
+
+    This effect 'spins' the qubit and brings it in or out phase
+    when compared to other qubits.   This has little effect
+    on the |0> and |1> states, but can change the phase (complex sign)
+    of states in superpositions.
+
+    A full phase is the same as a rotation around the Z axis, or a Z gate.
```
smitsanghavi, line 65:
```suggestion
    A full phase is the same as a 180° rotation around the Z axis, or a Z gate.
```

smitsanghavi (approved):

PR #82, opened 2023-03-28
Formatting to new version of black.

dstrain115:
- Reformatting changes only, done by running black on the quantum_rpg directory.


smitsanghavi:
How about running it on the whole of unitary at once? Just so that everything is consistent and at the same formatter.

smitsanghavi (approved):

PR #84, opened 2023-04-24
Add tictactoe main function and gameplay

khaledallen:
Creates a CLI for playing quantum tic tac toe to demonstrate the game.

### DIFF
--- a/unitary/examples/tictactoe/enums.py
+++ b/unitary/examples/tictactoe/enums.py
@@ -13,8 +13,9 @@
 # limitations under the License.
 #
 import enum
+from enum import StrEnum
 
-class GameMoves(enum.Enum):
+class GameMoves(StrEnum):
     EXIT = 0
     MAP = 1
     HELP = 2
--- a/unitary/examples/tictactoe/enums.py
+++ b/unitary/examples/tictactoe/enums.py
@@ -13,9 +13,8 @@
 # limitations under the License.
 #
 import enum
-from enum import StrEnum
 
-class GameMoves(StrEnum):
+class GameMoves(enum.Enum):
     EXIT = 0
     MAP = 1
     HELP = 2
--- a/unitary/examples/tictactoe/enums.py
+++ b/unitary/examples/tictactoe/enums.py
@@ -14,6 +14,10 @@
 #
 import enum
 
+class GameMoves(enum.Enum):
+    EXIT = 0
+    MAP = 1
+    HELP = 2
 
 class TicTacSquare(enum.Enum):
     EMPTY = 0
--- a/unitary/examples/tictactoe/enums.py
+++ b/unitary/examples/tictactoe/enums.py
@@ -15,9 +15,9 @@
 import enum
 
 class GameMoves(enum.Enum):
-    EXIT = 0
-    MAP = 1
-    HELP = 2
+    EXIT = 'exit'
+    MAP = 'map'
+    HELP = 'help'
 
 class TicTacSquare(enum.Enum):
     EMPTY = 0
--- a/unitary/examples/tictactoe/enums.py
+++ b/unitary/examples/tictactoe/enums.py
@@ -14,10 +14,12 @@
 #
 import enum
 
+
 class GameMoves(enum.Enum):
-    EXIT = 'exit'
-    MAP = 'map'
-    HELP = 'help'
+    EXIT = "exit"
+    MAP = "map"
+    HELP = "help"
+
 
 class TicTacSquare(enum.Enum):
     EMPTY = 0
--- a/unitary/examples/tictactoe/tic_tac_toe.py
+++ b/unitary/examples/tictactoe/tic_tac_toe.py
@@ -244,6 +244,23 @@ class GameInterface:
             for testing.
 
     """
+
+    help_text = """
+    You can enter:
+    - 1 character from [abcdefghi] to place a mark in the corresponding square (eg "a")
+    - 2 characters from [abcdefghi] to place a split mark in corresponding squares (eg "bd")
+    - "map": show board map
+    - "exit" to quit
+    """
+
+    board_map = """
+        a | b | c
+        -----------
+        d | e | f
+        -----------
+        g | h | i
+        """
+
     def __init__(self, game: TicTacToe, file: io.IOBase = sys.stdout):
         self.game = game
         self.file = file
@@ -266,15 +283,11 @@ class GameInterface:
         in which case this function hands the move off to the TicTacToe instance,
         or one of the GameMoves enums (GameMoves.MAP, GameMoves.EXIT, GameMoves.HELP),
         which prevent the game loop from alternating to the next player.
-
-        Raises:
-            ValueError if it is still the player's move.
         """
         move = self.get_move()
-        print(move)
 
         if move == GameMoves.MAP.name.lower():
-            print(self.print_board_map(), file=self.file)
+            print(self.board_map, file=self.file)
             print("Still your move.", file=self.file)
             return
         if move == GameMoves.EXIT.name.lower():
@@ -282,7 +295,7 @@ class GameInterface:
             print("Goodbye!", file=self.file)
             return
         if move == GameMoves.HELP.name.lower():
-            print(self.print_help(), file=self.file)
+            print(self.help_text, file=self.file)
             print("Still your move.", file=self.file)
             return
 
@@ -299,7 +312,7 @@ class GameInterface:
         Welcome to quantum tic tac toe!
         Here is the board:
         """
-        message += self.print_board_map()
+        message += self.board_map
         return message
 
     def play(self) -> None:
@@ -310,29 +323,13 @@ class GameInterface:
         """
         print(self.print_welcome(), file=self.file)
         while self.game.result() == TicTacResult.UNFINISHED and not self.player_quit: 
-            self.player_move()
+            try:
+                self.player_move()
+            except:
+                print('Catch error')
+
         print(self.game.result(), file=self.file)
     
-    def print_help(self) -> str:
-        """Print the available moves that player can take."""
-        return """
-    You can enter:
-    - 1 character from [abcdefghi] to place a mark in the corresponding square (eg "a")
-    - 2 characters from [abcdefghi] to place a split mark in corresponding squares (eg "bd")
-    - "map": show board map
-    - "exit" to quit
-    """
-
-    def print_board_map(self) -> str:
-        """Print the mapping of letters to board spaces for player move reference."""
-        return """
-        a | b | c
-        -----------
-        d | e | f
-        -----------
-        g | h | i
-        """
-
     def print_board(self) -> str:
         """Returns the TicTacToe board in ASCII form."""
         results = self.game.board.peek(count=100)
--- a/unitary/examples/tictactoe/tic_tac_toe.py
+++ b/unitary/examples/tictactoe/tic_tac_toe.py
@@ -265,7 +265,7 @@ class Game:
         self.game = game
         self.file = file
         self.player = "X"
-        self.playerQuit = False
+        self.player_quit = False
 
     def player_move(self) -> None: 
         """
@@ -284,7 +284,7 @@ class Game:
             print(self.print_board_map(), file=self.file)
             raise ValueError("Still your move.")
         if move == GameMoves.EXIT.name:
-            self.playerQuit = True
+            self.player_quit = True
             raise ValueError("Goodbye!", file=self.file)
         if move == GameMoves.HELP.name:
             print(self.print_help())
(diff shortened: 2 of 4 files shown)


unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -244,3 +244,71 @@ def print(self) -> str:
...
+
+    def setup(self):
+        print("Welcome to quantum tic tac toe!")
+        print("Here is the board:")
+        self.printBoardMap()
+        self.player = "X"
+        self.playerQuit = False
+
+    def handleMove(self, move): 
+        if move == 'map':
```
dstrain115, line 260: Should we change move to an enum so that the list of commands is clear?

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -244,3 +244,71 @@ def print(self) -> str:
...
+class Game:
+    def __init__(self, game):
+        self.game = game
+
+    def setup(self):
+        print("Welcome to quantum tic tac toe!")
+        print("Here is the board:")
+        self.printBoardMap()
+        self.player = "X"
+        self.playerQuit = False
```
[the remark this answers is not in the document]
khaledallen: I also realized the setup function was unnecessary so I removed that and put the setup steps into the play function.

unitary/examples/tictactoe/tic_tac_toe.py
khaledallen: Done.

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -244,3 +244,71 @@ def print(self) -> str:
             if idx in [2, 5, 8] and row != 2:
                 output += "--------------------------\n"
         return output
+
+class Game:
+    def __init__(self, game):
+        self.game = game
+
+    def setup(self):
+        print("Welcome to quantum tic tac toe!")
```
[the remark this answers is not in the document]
khaledallen: Done. I am going to review this pattern a bit more and may make some additional changes to the helper functions.

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -244,3 +244,71 @@ def print(self) -> str:
             if idx in [2, 5, 8] and row != 2:
                 output += "--------------------------\n"
         return output
+
+class Game:
+    def __init__(self, game):
```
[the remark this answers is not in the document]
khaledallen: Done

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -244,3 +244,71 @@ def print(self) -> str:
             if idx in [2, 5, 8] and row != 2:
                 output += "--------------------------\n"
         return output
+
+class Game:
```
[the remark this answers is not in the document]
khaledallen: Done.

unitary/examples/tictactoe/tic_tac_toe.py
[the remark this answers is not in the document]
khaledallen: I went through a had the helper print functions follow the same pattern as those in the TicTacToe class.

khaledallen:
Made a few more changes and correct the implementation with the enum to match the user input string with the enum name (rather than using StrEnum, which was proving finicky). Also went through and made sure the quotes were all consistent.

khaledallen:
I am thinking it might also make sense to move the TicTacToe.print() function into the Game class, since that isn't really game logic but is just for user-interaction. What do you think?

unitary/examples/tictactoe/enums.py
```diff
@@ -14,6 +14,10 @@
 #
 import enum
 
+class GameMoves(enum.Enum):
+    EXIT = 0
```
dstrain115, line 18: You should be able to do:

```
class GameMoves(enum.Enum):
    EXIT = 'exit'
    MAP = 'map'
    HELP = 'help'
```

Then, below you can do `GameMoves.EXIT.value` rather than doing string manipulation each time.

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -227,9 +229,113 @@ def sample(self, count: int = 1) -> List[str]:
...
+        A move can be a one or two letter string within the set [abcdefghi],
+        in which case this function hands the move off to the TicTacToe instance,
+        or one of the GameMoves enums (GameMoves.MAP, GameMoves.EXIT, GameMoves.HELP),
+        which prevent the game loop from alternating to the next player.
+
+        Raises:
+            ValueError if it is still the player's move.
+        """
+        move = self.get_move()
+        print(move)
```
dstrain115, line 274: Should this be `print(move, file=self.file)`

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -227,9 +229,113 @@ def sample(self, count: int = 1) -> List[str]:
...
+    
+    def print_help(self) -> str:
+        """Print the available moves that player can take."""
+        return """
+    You can enter:
+    - 1 character from [abcdefghi] to place a mark in the corresponding square (eg "a")
+    - 2 characters from [abcdefghi] to place a split mark in corresponding squares (eg "bd")
+    - "map": show board map
+    - "exit" to quit
+    """
```
dstrain115, line 324: Maybe this should be a constant instead of a function.

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -227,9 +229,113 @@ def sample(self, count: int = 1) -> List[str]:
...
+
+    def print_board_map(self) -> str:
+        """Print the mapping of letters to board spaces for player move reference."""
+        return """
+        a | b | c
+        -----------
+        d | e | f
+        -----------
+        g | h | i
+        """
```
dstrain115, line 334: Same here.  This should be a constant.

unitary/examples/tictactoe/tic_tac_toe_test.py
```diff
@@ -142,16 +144,107 @@ def test_rulesets(run_on_hardware):
...
+        d | e | f
+        -----------
+        g | h | i
+        """
+    )
+
+def test_help():
+    original_get_move = tictactoe.GameInterface.get_move
+    tictactoe.GameInterface.get_move = lambda _: 'help'
+    test_file = open("test.txt", "w")
```
dstrain115, line 167: I am not sure this is a great idea to open a file here.   If you are going to do this, you should use the pytest fixture or a function to make a temp file.  However, you should be able to use `io.StringIO()` instead.

See:
https://github.com/quantumlib/unitary/blob/main/unitary/examples/quantum_rpg/battle_test.py#L8

unitary/examples/tictactoe/tic_tac_toe_test.py
```diff
@@ -142,16 +144,107 @@ def test_rulesets(run_on_hardware):
...
+        -----------
+        g | h | i
+        
+Still your move.
+"""
+    )
+    assert (game.player == 'X')
+
+def test_exit():
+    original_get_move = tictactoe.GameInterface.get_move
```
dstrain115, line 214: Yikes.  This is super hacky.  Don't do this.  

Either mock get_move or modify it to provide a an input function as an argument, such as this (not quite sure if this exactly works as written but something like it):

```
    def get_move(self, input_fn: Callable[str,str]=input) -> str: 
        return input_fn(f"Player {self.player} to move (\"help\" for help): ")
```

unitary/examples/tictactoe/tic_tac_toe.py
khaledallen: True. Hold over from when these actually did the printing.

unitary/examples/tictactoe/enums.py
khaledallen: This is what I had tried initially, but `GameMoves.EXIT.value` returns `0`, so it doesn't work. I did try using StrEnum as well, with similarly frustrating results.

unitary/examples/tictactoe/tic_tac_toe.py
khaledallen: No, that was a testing artifact.

unitary/examples/tictactoe/tic_tac_toe_test.py
khaledallen: That's what I figured. I had tried using `@patch` decorators, but `MagicMock` ended up being much simpler.

unitary/examples/tictactoe/tic_tac_toe_test.py
khaledallen: Done

unitary/examples/tictactoe/enums.py
dstrain115: This is from an ipython I am running in my virtualenv:

```
In [2]: import enum

In [3]: class GameMoves(enum.Enum):
   ...:     EXIT = 'exit'
   ...:     MAP = 'map'
   ...:     HELP = 'help'
   ...: 

In [4]: GameMoves.EXIT.value
Out[4]: 'exit'
```

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -227,9 +229,110 @@ def sample(self, count: int = 1) -> List[str]:
...
+
+    help_text = """
+    You can enter:
+    - 1 character from [abcdefghi] to place a mark in the corresponding square (eg "a")
+    - 2 characters from [abcdefghi] to place a split mark in corresponding squares (eg "bd")
+    - "map": show board map
+    - "exit" to quit
+    """
+
+    board_map = """
```
dstrain115, line 256: Typically global constants are capitalized, such as BOARD_MAP (or _BOARD_MAP if it should be package private).  Also, I would put this above GameInterface, as I think it is now a member attribute when defined this way.

See https://google.github.io/styleguide/pyguide.html#316-naming

unitary/examples/tictactoe/tic_tac_toe.py
```diff
@@ -227,9 +229,110 @@ def sample(self, count: int = 1) -> List[str]:
...
+        Run the game loop, requesting player moves, alternating players, until
+        the TicTacToe instance reports that the game ends with a winner or a tie
+        or one of the players has quit.
+        """
+        print(self.print_welcome(), file=self.file)
+        while self.game.result() == TicTacResult.UNFINISHED and not self.player_quit: 
+            try:
+                self.player_move()
+            except:
+                print('Catch error')
```
dstrain115, line 329: This needs a better error message.   This should probably also be printing to self.file?

Also, why are we catching all errors indiscriminately?  Do we expect errors here?

dstrain115 (commented):
Getting close.  A few last comments.

unitary/examples/tictactoe/enums.py
khaledallen: I don't know what to say. It works when I run a python interpreter in the terminal and do exactly what you posted above. It doesn't work in the module. I added a `print(GameMoves.EXIT.value)` and it prints `0` even if I define the enum as you have above.

I added
```
        move = self.get_move()
        print('GameMoves.EXIT name and value')
        print(GameMoves.EXIT.name)
        print(GameMoves.EXIT.value)
```
to the `get_move()` function.

The output is:
```

        Welcome to quantum tic tac toe!
        Here is the board:
        
        a | b | c
        -----------
        d | e | f
        -----------
        g | h | i
        
Player X to move ("help" for help): help
GameMoves.EXIT name and value
EXIT
0
Catch error
```

That said, the tests pass with the enum using strings so...I'm guessing this is a version problem.

unitary/examples/tictactoe/tic_tac_toe.py
khaledallen: Oops, that was still WIP. Fixed to only catch ValueErrors from the TicTacToe class.

unitary/examples/tictactoe/tic_tac_toe_test.py
```diff
@@ -142,16 +145,87 @@ def test_rulesets(run_on_hardware):
...
+        a | b | c
+        -----------
+        d | e | f
+        -----------
+        g | h | i
+        
+Still your move.
+"""
+    )
+    assert (game.player == 'X')
```
dstrain115, line 201: nit, you don't need these parentheses.  Same below.

unitary/examples/tictactoe/tic_tac_toe_test.py
```diff
@@ -142,16 +145,87 @@ def test_rulesets(run_on_hardware):
...
+
+def test_exit():
+    output = io.StringIO()
+    board = tictactoe.TicTacToe()
+    game = tictactoe.GameInterface(board, file=output)
+    game.get_move = MagicMock(return_value='exit')
+    game.player_move()
+
+    assert (output.getvalue() == "Goodbye!\n"
```
dstrain115, line 210: Weird formatting and unnecesssary parentheses.

dstrain115 (approved):

dstrain115:
Thanks for addressing all the comments!

PR #88, opened 2023-05-24
Add a title screen for quantum RPG

dstrain115:
- Because every project needs more ASCII art.


unitary/examples/quantum_rpg/main_loop.py
```diff
@@ -40,6 +41,10 @@ def __init__(
         self.file = file
         self.party = party
 
+
+    def print_title_screen(self):
```
smitsanghavi, line 45: Should this be called from `__init__` or is the caller supposed to explicitly print the title?

dstrain115: I'd like to make it explicit, so that we can control the flow better.

PR #91, opened 2023-05-29
Initial version of having XP in Quantum RPG

dstrain115:
This PR introduces functions for having XP associated with encounters and being able to award it to different qaracters.

XP are quantum effects (gates) that can be added to the qaracter sheet (e.g. circuit).  After a battle is successfully won, the XP is gained and added to the qaracter.

### DIFF
--- a/unitary/examples/quantum_rpg/xp_utils.py
+++ b/unitary/examples/quantum_rpg/xp_utils.py
@@ -1,0 +1,84 @@
+# Copyright 2023 The Unitary Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List, Optional, Sequence
+
+import io
+import random
+import sys
+
+import unitary.alpha as alpha
+import unitary.examples.quantum_rpg.qaracter as qaracter
+
+
+class EncounterXp:
+    """What experience (XP) is possible from an encounter."""
+
+    def __init__(
+        self,
+        xp_choices: Sequence[Sequence[alpha.QuantumEffect]],
+        weights: Optional[Sequence[float]] = None,
+    ):
+        self.xp_choices = list(xp_choices)
+        if not weights:
+            self.weights = [1] * len(self.xp_choices)
+        else:
+            self.weights = list(weights)
+        if len(self.weights) != len(self.xp_choices):
+            raise ValueError(
+                "Weighted probabilities must match the length of the gate XP"
+            )
+
+    def choose(self) -> List[alpha.QuantumEffect]:
+        """Choose an XP possibility from the list of possible XP choices."""
+        return list(random.choices(self.xp_choices, weights=self.weights)[0])
+
+
+def award_xp(
+    party: Sequence[qaracter.Qaracter],
+    xp: Optional[EncounterXp],
+    user_input: Optional[Sequence[str]],
+    file: io.IOBase = sys.stdout,
+):
+    """Promopt user to choose qaracter to award xp to."""
+    if not xp:
+        return
+    if user_input is not None:
+        user_input = iter(user_input)
+        get_user_input = lambda _: next(user_input)
+    else:
+        get_user_input = input
+    print("You have been awarded XP!", file=file)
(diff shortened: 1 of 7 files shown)


unitary/examples/quantum_rpg/xp_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    file: io.IOBase = sys.stdout,
+):
+    """Promopt user to choose qaracter to award xp to."""
+    if not xp:
+        return
+    if user_input is not None:
+        user_input = iter(user_input)
+        get_user_input = lambda _: next(user_input)
+    else:
+        get_user_input = input
```
smitsanghavi, line 61: Just approved #87. input helper should be usable here and in the qaracter selector below once merged

unitary/examples/quantum_rpg/xp_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        return list(random.choices(self.xp_choices, weights=self.weights)[0])
+
+
+def award_xp(
+    party: Sequence[qaracter.Qaracter],
+    xp: Optional[EncounterXp],
+    user_input: Optional[Sequence[str]],
+    file: io.IOBase = sys.stdout,
+):
+    """Promopt user to choose qaracter to award xp to."""
```
smitsanghavi, line 54:
```suggestion
    """Prompt user to choose a qaracter to award XP to."""
```

unitary/examples/quantum_rpg/xp_utils.py
```diff
@@ -0,0 +1,84 @@
...
+            print(f"{idx+1}) {qar.name}", file=file)
+        qar_choice = int(get_user_input(">"))
+        qar = party[qar_choice - 1]
+        print("Current qaracter sheet:", file=file)
+        print(qar.circuit, file=file)
+        qubit_list = list(qar.active_qubits())
+        qubit_choices = []
+        num_qubits = effect.num_objects() or 1
+        for qubit_num in range(num_qubits):
+            print(f"Choose qubit for {effect}:", file=file)
```
smitsanghavi, line 79:
```suggestion
            print(f"Choose qubit {qubit_num} for {effect}:", file=file)
```

unitary/examples/quantum_rpg/xp_utils.py
```diff
@@ -0,0 +1,84 @@
...
+            for idx, q in enumerate(qubit_list):
+                print(f"{idx+1}) {q}", file=file)
+            choice = int(get_user_input(">")) - 1
+            effect(qar.get_hp(qubit_list[choice]))
```
smitsanghavi, line 83: This won't work for multi-qubit effects, right? Let's also have a multi-qubit effect in the tests.

smitsanghavi (changes requested):

unitary/examples/quantum_rpg/xp_utils.py
dstrain115: No, it certainly will not.  Good catch.  Also added a multi-qubit test.

unitary/examples/quantum_rpg/xp_utils.py
dstrain115: The disadvantage of splitting PRs into nice small chunks.

unitary/examples/quantum_rpg/xp_utils.py
```diff
@@ -0,0 +1,86 @@
...
+        print(f"Choose the qaracter to add the {effect} to:", file=file)
+        for idx, qar in enumerate(party):
+            print(f"{idx+1}) {qar.name}", file=file)
+        qar_choice = input_helpers.get_user_input_number(
+            get_user_input, ">", len(party)
+        )
+        qar = party[qar_choice - 1]
+        print("Current qaracter sheet:", file=file)
+        print(qar.circuit, file=file)
+        qubit_list = list(qar.active_qubits())
```
smitsanghavi, line 74: Let's validate and raise error if `len(qubit_list) < effect.num_objects()`? Maybe it would be better UX to only enumerate the eligible qaracters for the choice above instead of the whole party.

smitsanghavi (approved):

losos0:
After a battle you receive a specific gate that you can add to the circuit? Sounds more like loot than XP.

dstrain115:
> After a battle you receive a specific gate that you can add to the circuit? Sounds more like loot than XP.

Perhaps, but the XP goes with the nomenclature that the circuit is the character sheet, the # of qubits is the level, etc.

unitary/examples/quantum_rpg/xp_utils.py
smitsanghavi: @dstrain115 FYI. Sorry I forgot to mark this as a change request.

dstrain115: Yeah, I added a check and test for it.
See https://github.com/quantumlib/unitary/blob/main/unitary/examples/quantum_rpg/xp_utils_test.py#L92

PR #92, opened 2023-05-31
Add items to Quantum RPG

dstrain115:
- Add items you can interact with in Quantum RPG.
- These will be useful for signs, helpful NPCS, and puzzles.
- Right now, they can only print out a string.

Follow-up PRs will hook in custom functions for more complicated puzzles.  For right now, there is a untyped 'Callable' as a placeholder for such customizations until I figure out what the function signature should be (or maybe it will be a class, like ItemAction).

### DIFF
--- a/unitary/examples/quantum_rpg/main_loop.py
+++ b/unitary/examples/quantum_rpg/main_loop.py
@@ -97,11 +97,10 @@ class MainLoop:
             if cmd is not None:
                 self.world.move(cmd)
                 continue
-            words = current_input.split()
-            action = self.world.current_location.get_action(words[0])
+            action = self.world.current_location.get_action(current_input)
             if action is not None:
                 if isinstance(action, str):
-                    print(action + "\n", file=self.file)
+                    print(action, file=self.file)
                 print_room_description = False
                 # TODO: actions that are 'callables'
                 continue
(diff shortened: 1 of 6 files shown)


unitary/examples/quantum_rpg/main_loop_test.py
```diff
@@ -77,7 +84,7 @@ def test_do_simple_move() -> None:
     output = io.StringIO()
     c = classes.Analyst("Mensing")
     loop = main_loop.MainLoop([c], world.World(EXAMPLE_WORLD), output)
-    loop.loop(user_input=["e", "w", "quit"])
+    loop.loop(user_input=["e", "read sign", "w", "quit"])
```
losos0, line 87: The second word in "read sign" intentionally doesn't matter?

unitary/examples/quantum_rpg/main_loop.py
```diff
@@ -92,6 +97,14 @@ def loop(self, user_input: Optional[Sequence[str]] = None) -> None:
             if cmd is not None:
                 self.world.move(cmd)
                 continue
+            words = current_input.split()
+            action = self.world.current_location.get_action(words[0])
```
losos0, line 101: This appears to throw on an empty line

dstrain115: Removed.

losos0: There is still the same problem; the code that throws on an empty line is just in another function.

LGTM other than that

dstrain115: Fixed, with test.  I misunderstood what you meant at first.

PR #96, opened 2023-06-09
Quantum RPG: Allow multi-qubit effects in qaracters

dstrain115:
- Currently, add_quantum_effect only takes one qubit.
- This allows you to pass a variable amount of qubits into the function so that we can add two-qubit gates.


losos0 (commented):
LGTM

PR #98, opened 2023-06-11
Add copy to QuantumWorld

dstrain115:
- Adds a way to copy the state of a QuantumWorld.

### DIFF
--- a/unitary/alpha/quantum_world_test.py
+++ b/unitary/alpha/quantum_world_test.py
@@ -285,10 +285,13 @@ def test_undo(simulator, compile_to_qubits):
 
 
 @pytest.mark.parametrize("compile_to_qubits", [False, True])
-def test_copy(compile_to_qubits):
+@pytest.mark.parametrize("simulator", [cirq.Simulator, alpha.SparseSimulator])
+def test_copy(simulator, compile_to_qubits):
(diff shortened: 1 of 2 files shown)


unitary/alpha/quantum_world_test.py
```diff
@@ -284,6 +284,50 @@ def test_undo(simulator, compile_to_qubits):
     assert all(result[0] == Light.GREEN for result in results)
 
 
+def test_copy():
```
smitsanghavi, line 287: Can you also add the `compile_to_qubits` parameterization to ensure that copying does not break that somehow?

smitsanghavi (changes requested):

unitary/alpha/quantum_world_test.py
dstrain115: Done.

smitsanghavi: Still need to initialize the `QuantumWorld` with the param.

dstrain115: Oh yeah, of course I do.  Thanks for all the reviews!

smitsanghavi (approved):

unitary/alpha/quantum_world_test.py
smitsanghavi: No problemo!

PR #100, opened 2023-06-14
[Quantum RPG] Add functions to items

dstrain115:
- This allows items to have functions that are callable and can modify game state.
- These will be useful for puzzles and for things that trigger new qaracters to the party, etc.


smitsanghavi (approved):
Neat!

PR #107, opened 2023-06-28
Initial type checking fixes 

dstrain115:
- In preparation ofadding type checking to unitary,
fix errors reported by mypy.
- Mainly this is a lot of type casting and checking for None values.
- This is about half the type errors in unitary.

### DIFF
--- a/unitary/examples/quantum_rpg/input_helpers.py
+++ b/unitary/examples/quantum_rpg/input_helpers.py
@@ -8,7 +8,6 @@ _USER_INPUT = Callable[[str], str]
 _INVALID_MESSAGE = "Invalid number selected."
 
 
-
 def get_user_input_function(user_input: Optional[Sequence[str]] = None) -> _USER_INPUT:
     """Returns a lambda for getting user input.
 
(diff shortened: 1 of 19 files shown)


unitary/quantum_chess/circuit_transformer.py
```diff
@@ -76,7 +76,9 @@ def qubits_within(
             c += self.qubits_within(depth - 1, qubit + diff, qubit_list, visited)
         return c
 
-    def find_start_qubit(self, qubit_list: List[cirq.Qid], depth=3) -> cirq.GridQubit:
+    def find_start_qubit(
+        self, qubit_list: Iterable[cirq.Qid], depth=3
```
losos0, line 80: `Iterable` is too loose since you need to be able to iterate it more than once. `Sequence` I guess

unitary/examples/quantum_rpg/input_helpers.py
```diff
@@ -28,7 +27,7 @@ def get_user_input_number(
     message: Optional[str] = "",
```
losos0, line 27: Does it really need to be optional?

losos0 (approved):

PR #113, opened 2023-07-07
[Quantum RPG] Sorry, you have died

dstrain115:
- Add a final state for quantum RPG if you lose.
- Also fix a bug where we were not properly copying health status.


cognigami (approved):
Hilarious!

PR #116, opened 2023-07-14
[Quantum RPG] Add puzzle for Hadamard Hills

dstrain115:
- Adds lasers for Alice and Bob to communicate via BB84.  Will be used for the puzzle for the
second zone.

### DIFF
--- a/unitary/examples/quantum_rpg/bb84.py
+++ b/unitary/examples/quantum_rpg/bb84.py
@@ -1,0 +1,211 @@
+# Copyright 2023 The Unitary Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


unitary/examples/quantum_rpg/bb84.py
```diff
@@ -0,0 +1,211 @@
...
+            assert alice[idx + 1] == bob[idx + 1], "Eve has measured a qubit!"
+            key += alice[idx + 1]
+    word = ""
+    for idx in range(0, len(key), 8):
+        word += chr(int(key[idx : idx + 8], 2))
+    return word
+
+
+def _gen_keys(state: GameState):
+    if "akey" not in state.state_dict:
```
smitsanghavi, line 93: `akey` is never set anywhere. Should this be `alice`/`bob`?

dstrain115: Fixed, but in #118 
This PR got screwed up somehow.

PR #119, opened 2023-07-15
[Quantum RPG] Add regular expressions to item parsing

dstrain115:
- This will allow "TYPE blah" for puzzles.

### DIFF
--- a/unitary/examples/quantum_rpg/item.py
+++ b/unitary/examples/quantum_rpg/item.py
@@ -1,4 +1,5 @@
 from typing import Any, Callable, Optional, Sequence, Tuple, TYPE_CHECKING, Union
+import re
 
 import unitary.examples.quantum_rpg.game_state as game_state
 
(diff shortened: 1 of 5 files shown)


unitary/examples/quantum_rpg/item.py
```diff
@@ -58,8 +59,12 @@ def get_action(self, user_input: str) -> Optional[ITEM_ACTION_TYPE]:
             if isinstance(targets, str):
                 targets = [targets]
             if keyword == keywords or keyword in keywords:
-                if not targets or target in targets:
-                    return action
+                if isinstance(targets, re.Pattern):
```
smitsanghavi, line 62: Would be more general to allow a sequence of patterns or a mix of patterns and literals in `targets`. Won't add too much complexity to the code from what I see.

smitsanghavi (approved):

PR #120, opened 2023-07-15
[Quantum RPG] Gaining a level

dstrain115:
- Once the number of operations on each qubit exceeds your level, you will gain another level (and HP)

### DIFF
--- a/unitary/examples/quantum_rpg/xp_utils.py
+++ b/unitary/examples/quantum_rpg/xp_utils.py
@@ -48,6 +48,11 @@ class EncounterXp:
 
 
 def is_ready_for_next_level(qar: qaracter.Qaracter) -> bool:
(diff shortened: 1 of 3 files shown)


unitary/examples/quantum_rpg/xp_utils.py
```diff
@@ -47,6 +47,18 @@ def choose(self) -> List[alpha.QuantumEffect]:
         return list(random.choices(self.xp_choices, weights=self.weights)[0])
 
 
+def is_ready_for_next_level(qar: qaracter.Qaracter) -> bool:
```
smitsanghavi, line 50: Can you add a docstring with the same explanation as the PR description? I was confused what was happening here until I read that. Maybe also print some indication to the player on the criteria for leveling up.

smitsanghavi (approved):

PR #122, opened 2023-07-16
[Quantum RPG] Improve text formatting

dstrain115:
- Change status lines to fixed format to improve the alignment.  Otherwise, short or long names can throw off the alignment of the status line.
- Also adds a text message explaining what happened if the players or enemies escape the battle.


losos0:
LGTM

PR #124, opened 2023-07-20
[Quantum RPG] Some small text improvements

dstrain115:
- Get rid of spammy "X is DOWN" message for enemies and just skip it.
- Get rid of debug print statement
- Add a enemy number for more clear input.


unitary/examples/quantum_rpg/battle.py
```diff
@@ -113,7 +111,7 @@ def take_player_turn(self):
             self.print_screen()
             print(f"{current_player.name} turn:", file=self.file)
             if not current_player.is_active():
-                print(f"{current_player.name} is DOWN!", file=self.file)
+                print(f"{current_player.name} is DOWNi and cannot act!", file=self.file)
```
losos0, line 114: Typo `DOWNi`

LGTM otherwise

PR #127, opened 2023-07-24
[Quantum RPG] Quantopedia

dstrain115:
- Adds an in-game help option for a quantopedia.
- When you find the book or item that has info on the type of enemy you are facing, you can type 'q' and learn about it.

### DIFF
--- a/unitary/examples/quantum_rpg/game_state.py
+++ b/unitary/examples/quantum_rpg/game_state.py
@@ -43,17 +43,29 @@ class GameState:
         self.get_user_input = input_helpers.get_user_input_function(user_input)
         self.file = file
 
-    def set_quantopedia(self, num):
+    def set_quantopedia(self, num: int) -> None:
+        """Convenience method for setting a quantopedia bit.
+
+        The quantopedia is repesented by a bitstring in the
+        state_dict.  During a battle, if the bit for the enemy
+        NPC is set, then the entry for that type of NPC is displayed.
+        """
         if _QUANTOPEDIA_KEY not in self.state_dict:
             self.state_dict[_QUANTOPEDIA_KEY] = "0"
         quantopedia_num = int(self.state_dict[_QUANTOPEDIA_KEY]) | num
         self.state_dict[_QUANTOPEDIA_KEY] = str(quantopedia_num)
 
-    def has_quantopedia(self, num):
+    def has_quantopedia(self, num: int) -> bool:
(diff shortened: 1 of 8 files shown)


unitary/examples/quantum_rpg/game_state.py
```diff
@@ -42,6 +43,18 @@ def __init__(
...
+        if _QUANTOPEDIA_KEY not in self.state_dict:
+            self.state_dict[_QUANTOPEDIA_KEY] = "0"
+        quantopedia_num = int(self.state_dict[_QUANTOPEDIA_KEY]) | num
+        self.state_dict[_QUANTOPEDIA_KEY] = str(quantopedia_num)
+
+    def has_quantopedia(self, num):
+        if _QUANTOPEDIA_KEY not in self.state_dict:
+            return False
+        quantopedia_num = int(self.state_dict[_QUANTOPEDIA_KEY])
+        return (quantopedia_num // num) % 2 == 1
```
losos0, line 56: `quantopedia_num & num != 0`

unitary/examples/quantum_rpg/npcs.py
```diff
@@ -67,6 +67,14 @@ def npc_action(self, battle, **kwargs) -> str:
         action_choice = random.random()
         return self.act_on_enemy_qubit(enemy_qubit, action_choice, **kwargs)
 
+    def quantopedia_index(self) -> int:
+        """Bit to use for quantopedia.  Should be power of two."""
+        return 1
```
losos0, line 72: Maybe 0 would be a better default? (impossible to learn about it)

losos0 (commented):
I guess it's lacking a call to set_quantopedia to be usable, but what's written so far seems OK.

unitary/examples/quantum_rpg/game_state.py
dstrain115: Done.

unitary/examples/quantum_rpg/npcs.py
dstrain115: Done.  Also, good point, I have added docstrings all around and have also added constants for each bit.

dstrain115:
FYI, set_quantopedia is called when reading books in oxtail, 
[]see here](https://github.com/quantumlib/unitary/pull/127/files#diff-fe847231b3c299f4d33e2b2b16ffe286acbdc2ed62e01ffb1a76663478cb7c88R77)

PR #141, opened 2023-07-27
[Quantum RPG] Add unit test for Level 3 Qaracters

dstrain115:
- Add unit test for level 3 qaracters to make sure descriptions are
  there.
- Also, change help to "?" since "h" is used for Hadamard already.


cantwellc (approved):
lgtm

Issue #137, opened 2023-07-27
Quantum RPG: Show status when assigning new effects


cantwellc:
It would be helpful to see the circuits when deciding who to give an effect to.

cantwellc:
I know it shows it when you pick the character, but at that point you can't go back and select the other character.

PR #145, opened 2023-08-01
[Quantum RPG] Display qaracter sheet during XP

dstrain115:
Fixes: #137


losos0 (approved):

cognigami (approved):
Been a while since I've had to think in regex.

PR #148, opened 2023-08-26
Fix status display for qubits with no effects

losos0:
It is confusing how the status display doesn't show qubits that don't have any effects on them. To fix this, specify a qubit order for the diagram, which allows us to force unused qubits to be included.


dstrain115 (approved):
Thanks for helping improve Quantum RPG!

PR #158, opened 2023-09-29
change enum.Piece to add color and language

madcpf:
- Add Language and Color as new enum classes;
- Rename enum class Piece to be Type;
- In a new file define class Piece inheriting from QuantumObject.


unitary/examples/quantum_chinese_chess/enums.py
```diff
@@ -42,31 +47,38 @@ class MoveVariant(enum.Enum):
...
+    SOLDIER = ["s", "S", "兵", "卒"]
+    CANNON = ["c", "C", "炮", "砲"]
+    ROOK = ["r", "R", "车", "車"]
+    HORSE = ["h", "H", "马", "馬"]
+    ELEPHANT = ["e", "E", "象", "相"]
+    ADVISOR = ["a", "A", "士", "仕"]
+    GENERAL = ["g", "G", "将", "帥"]
+
+    @staticmethod
+    def type_of(c: str) -> Optional["Type"]:
```
dstrain115, line 74: This function is no longer correct, it won't handle the Chinese symbols, correct?

unitary/examples/quantum_chinese_chess/enums.py
```diff
@@ -42,31 +47,38 @@ class MoveVariant(enum.Enum):
...
+    """The four values are symbols corresponding to
+    English red
+    English black
+    Chinese red
+    Chinese black
+    """
 
-    def black_symbol(self) -> str:
-        return self.value
+    EMPTY = [".", ".", ".", "."]
```
dstrain115, line 64: I would probably use tuples instead of lists here, since you do not want them to be mutable.

unitary/examples/quantum_chinese_chess/piece.py
```diff
@@ -0,0 +1,45 @@
...
+)
+
+
+class Piece(QuantumObject):
+    def __init__(self, name: str, state: SquareState, type_: Type, color: Color):
+        QuantumObject.__init__(self, name, state)
+        self.type_ = type_
+        self.color = color
+
+    def symbol(self, lang: Language = Language.EN) -> str:
```
dstrain115, line 29: Is it worth adding color as an argument then moving this as part of Type?

madcpf: Thanks Doug. It's updated.

unitary/examples/quantum_chinese_chess/enums.py
madcpf: Thanks Doug. It's updated.

unitary/examples/quantum_chinese_chess/enums.py
madcpf: Thanks Doug. This method (currently) is only used for convert FEN symbols into board pieces. I will rewrite it if further needs merge.

unitary/examples/quantum_chinese_chess/enums.py
```diff
@@ -42,31 +47,56 @@ class MoveVariant(enum.Enum):
...
+            if color == Color.RED:
+                return type_.value[0]
+            elif color == Color.BLACK:
+                return type_.value[1]
+        elif lang == Language.ZH:  # Return Chinese symbols
+            if color == Color.RED:
+                return type_.value[2]
+            elif color == Color.BLACK:
+                return type_.value[3]
+        return "Unexpected combinations"
```
dstrain115, line 102: I would probably raise an ValueError rather than silently pass a weird string.

dstrain115 (approved):

unitary/examples/quantum_chinese_chess/enums.py
madcpf: Thanks Doug. It's updated.

PR #159, opened 2023-10-03
Add Board class

madcpf:
Add Board class with methods:
- load_fen() to initialize the board;
- print() to print out the board.


unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,85 @@
...
+    SquareState,
+    GameState,
+    Color,
+    Type,
+    Language,
+)
+from unitary.examples.quantum_chinese_chess.piece import Piece
+
+
+class Board:
```
dstrain115, line 25: Please add a docstring.

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,85 @@
...
+    def __init__(self):
+        self.load_fen()
+        self.king_locations = {"e0", "e9"}
+        self.current_state = GameState.CONTINUE
+
+    def load_fen(
```
dstrain115, line 31: Please add a docstring.  Possibly link to a definition of how FEN format works for quantum chess.

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,85 @@
...
+        self, fen: str = "RHEAKAEHR/9/1C5C1/P1P1P1P1P/9/9/p1p1p1p1p/1c5c1/9/rheakaehr"
```
dstrain115, line 32: Please make this string into a constant, perhaps indicating that it is the initial board position.

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,85 @@
...
+    ):
+        chess_board = {}
+        row_index = 9
+        for row in fen.split("/"):
+            col = ord("a")
+            for char in row:
+                # Add empty board pieces.
+                if "1" <= char <= "9":
+                    for i in range(int(char)):
+                        name = chr(col) + "%i" % row_index
```
dstrain115, line 42: The new way to do this is `name = f'{chr(col)}{row_index}'`.  Same below.

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,85 @@
...
+                    chess_board[name] = Piece(
+                        name, SquareState.OCCUPIED, piece_type, color
+                    )
+                    col += 1
+            row_index -= 1
+        self.board = alpha.QuantumWorld(chess_board.values())
+
+    def print(self, lang: Language = Language.EN):
+        num_rows = 10
+        board_string = ""
```
dstrain115, line 61: Since strings are immutable, this is recreating the string each time.  Consider making this board_string = [] and then board_string.append(...).  Then, at the end, `print(''.join(board_string))`

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,85 @@
...
+                    board_string += " "
+            # Print the row index on the right.
+            board_string += " %d\n" % row
+        board_string += " "
+        # Print the bottom line of col letters.
+        for col in "abcdefghi":
+            board_string += " %c" % col
+        print(board_string)
+
+    def flying_general(self) -> bool:
```
dstrain115, line 83: Maybe remove this until it is completed.

unitary/examples/quantum_chinese_chess/board_test.py
```diff
@@ -0,0 +1,33 @@
...
+from unitary.examples.quantum_chinese_chess.board import Board
+
+
+def test_init(capsys):
+    board = Board()
+    board.print()
+    output = capsys.readouterr()
+    assert (
+        output.out
+        == "  a b c d e f g h i\n0 r h e a k a e h r  0\n1 . . . . . . . . .  1\n2 . c . . . . . c .  2\n3 p . p . p . p . p  3\n4 . . . . . . . . .  4\n5 . . . . . . . . .  5\n6 P . P . P . P . P  6\n7 . C . . . . . C .  7\n8 . . . . . . . . .  8\n9 R H E A K A E H R  9\n  a b c d e f g h i\n"
```
dstrain115, line 25: Consider changing the following so it is more readable:

```
assert (
        output.out
        == """  a b c d e f g h i
0 r h e a k a e h r  0
1 . . . . . . . . .  1
2 . c . . . . . c .  2
3 p . p . p . p . p  3
4 . . . . . . . . .  4
5 . . . . . . . . .  5
6 P . P . P . P . P  6
7 . C . . . . . C .  7
8 . . . . . . . . .  8
9 R H E A K A E H R  9
  a b c d e f g h i"
)
```

unitary/examples/quantum_chinese_chess/enums.py
```diff
@@ -25,6 +25,11 @@ class SquareState(enum.Enum):
     OCCUPIED = 1
 
 
+class GameState(enum.Enum):
```
dstrain115, line 28: Again, I would just add this when you add the rule check.

unitary/examples/quantum_chinese_chess/board_test.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+
+    board.print(Language.ZH)
+    output = capsys.readouterr()
```
dstrain115, line 29: Optional:  Consider having Board.print() just return a string so you can test it easier.

unitary/examples/quantum_chinese_chess/board_test.py
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from unitary.examples.quantum_chinese_chess.enums import Language
+from unitary.examples.quantum_chinese_chess.board import Board
+
+
```
dstrain115, line 18: I would add a test for load_fen that is not the initial position.

madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/board_test.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/enums.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Done. Thanks!

unitary/examples/quantum_chinese_chess/board.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/board_test.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,109 @@
...
+                    color = Color.RED if char.isupper() else Color.BLACK
+                    chess_board[name] = Piece(
+                        name, SquareState.OCCUPIED, piece_type, color
+                    )
+                    col += 1
+            row_index -= 1
+        self.board = alpha.QuantumWorld(chess_board.values())
+        self.current_player = 0 if "w" in turns else 1
+
+    def to_str(self, lang: Language = Language.EN):
```
dstrain115, line 69: Add return type annotation `->str`

Optional:  Do you think this would be better as `def __str__` so that you could just `print(board)` and it would work?

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class Board:
+    """Board holds the assemble of all pieces. Each piece could be either in classical or quantum state."""
+
+    def __init__(self, fen: str = _INITIAL_FEN):
+        self.king_locations = []
+        self.load_fen(fen)
+
+    def load_fen(self, fen: str):
```
dstrain115, line 35: What do you think about doing something like this instead?

```
def __init__(self, board: alpha.QuantumWorld, current_player: int):
    ....

@classmethod
def from_fen(cls, fen: str) -> Board:
   ...
  return cls(board, current_player)
```

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -0,0 +1,109 @@
...
+                    if piece_type == Type.KING:
+                        self.king_locations.append(name)
+                    color = Color.RED if char.isupper() else Color.BLACK
+                    chess_board[name] = Piece(
+                        name, SquareState.OCCUPIED, piece_type, color
+                    )
+                    col += 1
+            row_index -= 1
+        self.board = alpha.QuantumWorld(chess_board.values())
+        self.current_player = 0 if "w" in turns else 1
```
dstrain115, line 67: Should current_player be a enums.Color instead?

dstrain115 (approved):
A few more comments then LGTM.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Thanks Doug. It's updated.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Thanks Doug. It's updated.

unitary/examples/quantum_chinese_chess/board.py
madcpf: I would keep it as int since in the main chess.py file (which will be in the next review) it would be more convenient to do 1-current_player to switch to the other player.

PR #163, opened 2023-10-11
Fix typos in conceptes.ipynb, where m0 should be m1.

madcpf:


dstrain115 (approved):
It seems like it could be either m0 or m1.  I don't see where it was actually measured, but I am fine with either one.

PR #166, opened 2023-10-23
[Quantum Chinese Chess] Add SplitJump and MergeJump 

madcpf:
Add SplitJump and MergeJump as new moves.


unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -176,3 +176,57 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
         target_0.reset(source_0)
         source_0.reset()
         return iter(())
+
+
+class SplitJump(QuantumEffect):
+    """SplitJump from source_0 to target_0 and target_1. The only accepted (default) move_variant is
+    - BASIC.
```
dstrain115, line 183: Probably some more detail in the docstring would be helpful.  This doesn't really tell you much.

Maybe explain which kinds of chinese chess moves this covers and how it is implemented?

unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -176,3 +176,57 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
...
+        return 3
+
+    def effect(self, *objects) -> Iterator[cirq.Operation]:
+        source_0, source_1, target_0 = objects
+        # Make the merge jump.
+        alpha.PhasedMove(-0.5)(source_0, target_0)
+        alpha.PhasedMove(-0.5)(source_0, target_0)
+        alpha.PhasedMove(-0.5)(source_1, target_0)
+        # Pass the classical properties of the source pieces to the target piece.
+        target_0.reset(source_0)
```
dstrain115, line 231: This semantic is kind of confusing.  I am not totally sure if this is correct.  Remind me what reset is supposed to do again?

unitary/examples/quantum_chinese_chess/move_test.py
```diff
@@ -261,3 +261,113 @@ def test_jump_basic():
     assert len(board_probabilities) == 2
     assert_fifty_fifty(board_probabilities, locations_to_bitboard(["d1"]))
     assert_fifty_fifty(board_probabilities, locations_to_bitboard(["a3"]))
+
+
+def test_split_jump_classical_source():
```
dstrain115, line 266: Do you think it is worth testing the is_entangled / reset functionality to make sure it is correct?

unitary/examples/quantum_chinese_chess/move.py
madcpf: It copies the "type", "color" and "is_entangled" properties from source_0 to target_0.

dstrain115: I see.  And we are assuming that type, color are the same for source_0 and source_1.

Couldn't is_entangled be different for source_0 and source_1 though?
Also, do we ever fix the type/color for the source squares?  Like, they may be empty now, right?

I don't think this logic is correct.

unitary/examples/quantum_chinese_chess/move.py
madcpf: Thanks Doug. It's added.

unitary/examples/quantum_chinese_chess/move_test.py
madcpf: Thanks Doug. It's added in the test.

unitary/examples/quantum_chinese_chess/move.py
madcpf: Hi Doug, yes, earlier in chess.py we've checked to make sure source_0 and source_1 have the same type, color, and is_entangled = true (maybe later we could support classical to merge with entangled piece).

The type/color/is_entangled for source squares are fixed through the method "update_board_by_sampling()" in chess.py, which will be called after each move. There are cases that they won't be reset to empty (see test cases for those scenarios).

The source piece for SplitJump will also be fixed through update_board_by_sampling().

unitary/examples/quantum_chinese_chess/move_test.py
```diff
@@ -261,3 +261,121 @@ def test_jump_basic():
...
+            locations_to_bitboard(["b3", "c2"]): 1.0 / 8,
+            locations_to_bitboard(["b2", "b3"]): 1.0 / 8,
+            locations_to_bitboard(["b2", "c2"]): 1.0 / 4,
+        },
+    )
+
+
+def test_merge_jump_imperfect_merge_scenario_3():
+    # Imperfect merge scenario 3
+    # This is a simplied version of the scenario above, where we unhook a3 and b3.
```
dstrain115, line 363: Consider changing the line comments here and above to docstrings with triple quotes.  I think they will show up in IDEs better.

dstrain115 (approved):

unitary/examples/quantum_chinese_chess/move_test.py
madcpf: Thanks Doug! It's updated.

PR #167, opened 2023-10-30
Add one function into QuantumWorld

madcpf:
- get_histogram_with_all_objects_as_key(), which creates histogram of the whole quantum world (or `objects` if specified) based on measurements (peeks) carried out;


unitary/alpha/quantum_world.py
```diff
@@ -484,6 +486,35 @@ def get_histogram(
                 histogram[idx][cast(int, result[idx])] += 1
         return histogram
 
+    def get_histogram_with_all_objects_as_key(
```
dstrain115, line 489: I think we might want to come up with a better name.  Maybe get_correlated_histogram?  @cognigami @cantwellc any suggestions?

unitary/alpha/quantum_world.py
```diff
@@ -526,6 +557,103 @@ def get_binary_probabilities(
...
+                        of the state vector.
+
+        Returns:
+            A list with one element for each object which contains the probability
+            for the event state!=0, which is the same as 1.0-Probability(state==0).
+        """
+        if objects is None:
+            objects = self.public_objects
+
+        simulator = cirq.Simulator()
```
dstrain115, line 579: I don't know that this is a good idea to use cirq.Simulator().  Shouldn't we use self.sampler?

unitary/alpha/quantum_world.py
```diff
@@ -526,6 +557,103 @@ def get_binary_probabilities(
             binary_probs.append(1 - one_probs[0])
         return binary_probs
 
+    # TODO(pengfeichen): make this function work for compile_to_qubits==True.
+    def get_binary_probabilities_from_state_vector(
```
dstrain115, line 561: I'm pretty uncomfortable adding this.  Nothing else in the alpha library assumes that we are using a simulator that can create a state vector.  Why do we need this anyway?  Is it because get_histogram_with_all_objects_as_key is too slow and / or needs too many samples?

unitary/alpha/quantum_world.py
madcpf: Thanks Doug. It's updated to use self.sampler.

unitary/alpha/quantum_world.py
madcpf: Hi Doug, I was adding this so that we have a choice to access the accurate/expected probability of quantum objects within a quantum world, while the current method is based on sampling and the the returned probability varies call by call.

I did not compare in terms of how fast this implementation vs sampling.

If you don't think it's useful, I can remove it. Thanks.

unitary/alpha/quantum_world.py
madcpf: Thanks Doug. It's updated. @cognigami @cantwellc please let me know if you have other suggestions. Thanks.

unitary/alpha/quantum_world.py
madcpf: Hi Doug, after some discussion this morning in the meeting, I've removed this util function. Thanks.

unitary/alpha/quantum_world.py
madcpf: Hi Doug, after some discussion this morning in the meeting, I've removed this util function. Thanks.

unitary/alpha/quantum_world.py
```diff
@@ -484,6 +486,35 @@ def get_histogram(
...
+        based on measurements (peeks) carried out. Comparing to get_histogram(),
+        this statistics contains entanglement information accross quantum objects.
+
+        Parameters:
+            objects:    List of QuantumObjects
+            count:      Number of measurements
+
+        Returns:
+            A dictionary, with the keys being each possible state of the whole quantum
+            world (or `objects` if specified) in terms of tuple, and the values being
```
dstrain115, line 502: Let's clarify that the key is a tuple of the results of each object in `objects` and in the order of that argument.

unitary/alpha/quantum_world_test.py
```diff
@@ -673,6 +677,10 @@ def test_get_histogram_and_get_probabilities_one_binary_qobject(
     assert len(histogram[0]) == 2
     assert histogram[0][0] > 10
     assert histogram[0][1] > 10
+    histogram = world.get_correlated_histogram()
+    assert len(histogram) == 2
+    assert histogram[(0,)] > 10
+    assert histogram[(1,)] > 10
```
dstrain115, line 683: Should we have a test where the correlated histogram actually captures some entanglement and is different than the normal histogram?

dstrain115 (approved):

unitary/alpha/quantum_world_test.py
madcpf: Thanks Doug! A new test is added.

unitary/alpha/quantum_world.py
madcpf: Thanks Doug. It's updated.

PR #168, opened 2023-10-30
[Quantum Chinese Chess] Update flying general check

madcpf:
- Update flying general check to include the cases where there are quantum pieces between two kings(aka generals).
- Update Move to not depend on Board to break the cyclic dependency.

### DIFF
--- a/unitary/examples/quantum_chinese_chess/board.py
+++ b/unitary/examples/quantum_chinese_chess/board.py
@@ -21,6 +21,7 @@ from unitary.examples.quantum_chinese_chess.enums import (
     Language,
 )
 from unitary.examples.quantum_chinese_chess.piece import Piece
+from unitary.examples.quantum_chinese_chess.move import Jump
 
 
 # The default initial state of the game.
--- a/unitary/examples/quantum_chinese_chess/board.py
+++ b/unitary/examples/quantum_chinese_chess/board.py
@@ -19,6 +19,7 @@ from unitary.examples.quantum_chinese_chess.enums import (
     Color,
     Type,
     Language,
+    MoveVariant,
 )
 from unitary.examples.quantum_chinese_chess.piece import Piece
 from unitary.examples.quantum_chinese_chess.move import Jump
--- a/unitary/examples/quantum_chinese_chess/board.py
+++ b/unitary/examples/quantum_chinese_chess/board.py
@@ -22,7 +22,7 @@ from unitary.examples.quantum_chinese_chess.enums import (
     MoveVariant,
 )
 from unitary.examples.quantum_chinese_chess.piece import Piece
-from unitary.examples.quantum_chinese_chess.move import *
+from unitary.examples.quantum_chinese_chess.move import Jump
 
 
 # The default initial state of the game.
--- a/unitary/examples/quantum_chinese_chess/board.py
+++ b/unitary/examples/quantum_chinese_chess/board.py
@@ -183,8 +183,8 @@ class Board:
         if len(quantum_pieces) == 0:
             # If there are no pieces between two KINGs, the check successes. Game ends.
             return True
-        # When there are quantum pieces in between, the check successes (and game ends)
-        # if there are at lease one occupied path piece.
+        # When there are only quantum pieces in between, the check successes (and game ends)
+        # if all of those pieces turn out to be empty.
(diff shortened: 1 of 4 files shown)


unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -15,23 +15,21 @@
...
     """The base class of all chess moves."""
 
     def __init__(
         self,
-        source: str,
-        target: str,
-        board: Board,
-        source2: Optional[str] = None,
-        target2: Optional[str] = None,
+        source: Piece,
```
dstrain115, line 29: I am not sure if you need a board, but just FYI, if board is only needed for type checking, you can do 

`board: 'Board',` (with quotes)
and add `from __future__ import annotations`
(or hide it behind a `if TYPE_CHECKING` call)

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -183,7 +185,33 @@ def flying_general_check(self) -> bool:
...
+        if could_capture:
+            # Force measure all path pieces to be empty.
+            for path_piece in control_objects:
+                self.board.force_measurement(path_piece, 0)
+                path_piece.reset()
+
+            # Let the general/king fly, i.e. the opposite king will capture the current king.
+            current_king = self.board[self.king_locations[self.current_player]]
+            oppsite_king = self.board[self.king_locations[1 - self.current_player]]
+            Jump(MoveVariant.CLASSICAL)(oppsite_king, current_king)
```
dstrain115, line 206: The game ends here, right?  So, is it necessary to actually move the king?  I guess there's no reason not to, but it seems unnecessary.

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -183,7 +185,33 @@ def flying_general_check(self) -> bool:
...
+            print("==== FLYING GENERAL ! ====")
+            return True
+        else:
+            # TODO(): we are leaving the path pieces unchanged in entangled state. Maybe
```
dstrain115, line 210: I think this is fine, right?  since we measured the ancilla, it's guaranteed that all states remaining on the board have some piece in between.

unitary/examples/quantum_chinese_chess/board_test.py
```diff
@@ -155,3 +163,32 @@ def test_flying_general_check():
...
+    world = board.board
+    alpha.PhasedSplit()(world["a3"], world["c3"], world["e3"])
+    world["e3"].is_entangled = True
+    alpha.PhasedSplit()(world["a4"], world["c4"], world["e4"])
+    world["e4"].is_entangled = True
+
+    result = board.flying_general_check()
+    # We check the ancilla to learn whether the general/king flies or not.
+    captured = world.post_selection[world["ancilla_ancilla_flying_general_check_0_0"]]
+    if captured:
```
dstrain115, line 182: It would be nice to know that each branch of the if statement happens. (That some of the time it is captured and some of the time it is not).

One way to get one branch deterministically would be to test this:
```
    board = set_board(["a3", "e0", "e9"])
    board.king_locations = ["e0", "e9"]
    board.current_player = 0  # i.e. RED
    alpha.PhasedSplit()(world["a3"], world["e3"], world["e5"])

```

unitary/examples/quantum_chinese_chess/move.py
madcpf: Thanks a lot Doug! This is new to me.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Hi Doug, yes the game ends here. I just chose to capture the other king so that later on we print the updated board as the last print of the game.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Thanks Doug. Yes I've updated the comment here.

unitary/examples/quantum_chinese_chess/board_test.py
madcpf: Thanks Doug. I just added this case as a separate test, where capture could not happen. But I cannot think of a way to deterministically make sure the capture happens.

dstrain115 (approved):

PR #172, opened 2023-11-30
[Quantum Chinese Chess] Implement more move types

madcpf:
Implement the following types of moves:
- slide
- split slide
- merge slide
- cannon fire

### DIFF
--- a/unitary/examples/quantum_chinese_chess/chess.py
+++ b/unitary/examples/quantum_chinese_chess/chess.py
@@ -21,7 +21,7 @@ from unitary.examples.quantum_chinese_chess.enums import (
     MoveType,
     MoveVariant,
 )
-from unitary.examples.quantum_chinese_chess.move import Jump
+from unitary.examples.quantum_chinese_chess.move import *
 import readline
 
 # List of accepable commands.
--- a/unitary/examples/quantum_chinese_chess/chess.py
+++ b/unitary/examples/quantum_chinese_chess/chess.py
@@ -21,7 +21,15 @@ from unitary.examples.quantum_chinese_chess.enums import (
     MoveType,
     MoveVariant,
 )
-from unitary.examples.quantum_chinese_chess.move import *
+from unitary.examples.quantum_chinese_chess.move import (
+    Jump,
+    SplitJump,
+    MergeJump,
+    Slide,
+    SplitSlide,
+    MergeSlide,
+    CannonFire,
+)
 import readline
 
 # List of accepable commands.
@@ -390,6 +398,7 @@ class QuantumChineseChess:
             print(move_type, " ", move_variant)
 
         # Apply the move accoding to its type.
+        # TODO(): using match...case... when python 3.11 satisfies the dependency.
         if move_type == MoveType.CLASSICAL:
             if source_0.type_ == Type.KING:
                 # Update the locations of KING.
--- a/unitary/examples/quantum_chinese_chess/chess.py
+++ b/unitary/examples/quantum_chinese_chess/chess.py
@@ -402,7 +402,18 @@ class QuantumChineseChess:
             Jump(move_variant)(source_0, target_0)
         elif move_type == MoveType.JUMP:
             Jump(move_variant)(source_0, target_0)
-        # TODO(): apply other move types.
+        elif move_type == MoveType.SLIDE:
+            Slide(quantum_pieces_0, move_variant)(source_0, target_0)
+        elif move_type == MoveType.SPLIT_JUMP:
+            SplitJump()(source_0, target_0, target_1)
+        elif move_type == MoveType.SPLIT_SLIDE:
+            SplitSlide(quantum_pieces_0, quantum_pieces_1)(source_0, target_0, target_1)
+        elif move_type == MoveType.MERGE_JUMP:
+            MergeJump()(source_0, source_1, target_0)
+        elif move_type == MoveType.MERGE_SLIDE:
+            MergeSlide(quantum_pieces_0, quantum_pieces_1)(source_0, source_1, target_0)
+        elif move_type == MoveType.CANNON_FIRE:
+            CannonFire(classical_pieces_0, quantum_pieces_0)(source_0, target_0)
 
     def next_move(self) -> Tuple[bool, str]:
         """Check if the player wants to exit or needs help message. Otherwise parse and apply the move.
--- a/unitary/examples/quantum_chinese_chess/move.py
+++ b/unitary/examples/quantum_chinese_chess/move.py
@@ -593,9 +593,7 @@ class CannonFire(QuantumEffect):
             for index, expect_occupied_path_piece in enumerate(quantum_path_pieces_0):
                 # TODO(): consider specific cases to save the ancilla.
                 # Add a new ancilla to indicate whether the fire could be made (value = 1 means it could).
-                capture_ancilla = world._add_ancilla(
-                    f"{expect_occupied_path_piece.name}"
-                )
+                capture_ancilla = world._add_ancilla(expect_occupied_path_piece.name)
                 # All other path pieces are expected to be empty to make the fire happen.
                 expect_empty_pieces = [
                     piece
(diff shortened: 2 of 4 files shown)


unitary/examples/quantum_chinese_chess/chess.py
```diff
@@ -402,7 +402,18 @@ def apply_move(self, str_to_parse: str) -> None:
             Jump(move_variant)(source_0, target_0)
         elif move_type == MoveType.JUMP:
             Jump(move_variant)(source_0, target_0)
-        # TODO(): apply other move types.
+        elif move_type == MoveType.SLIDE:
```
solodov, line 405: what version of python is this targetting?  With 3.11 this big if/else can be made much nicer with:

```py
match move_type:
    case MoveType.CLASSICAL:
        # code
    case MoveType.JUMP:
        # code
```

I think we'd also get lint checks of not all enum values are checked.

unitary/examples/quantum_chinese_chess/chess.py
```diff
@@ -21,7 +21,7 @@
     MoveType,
     MoveVariant,
 )
-from unitary.examples.quantum_chinese_chess.move import Jump
+from unitary.examples.quantum_chinese_chess.move import *
```
solodov, line 24: avoid star imports?

unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -217,3 +220,418 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
...
+        self.quantum_path_pieces_0 = quantum_path_pieces_0
+        self.move_variant = move_variant
+
+    def num_dimension(self) -> Optional[int]:
+        return 2
+
+    def num_objects(self) -> Optional[int]:
+        return 2
+
+    def effect(self, *objects) -> Iterator[cirq.Operation]:
```
solodov, line 249: Is there any expectation of type of values in *objects? perhaps add typing annotation?

unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -217,3 +220,418 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
...
+        source_0, target_0 = objects
```
solodov, line 250: this might be error-prone, the method signature says any number of objects, but this unpacking will break if not exactly two.

unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -217,3 +220,418 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
...
+        self.quantum_path_pieces_1 = quantum_path_pieces_1
+
+    def num_dimension(self) -> Optional[int]:
+        return 2
+
+    def num_objects(self) -> Optional[int]:
+        return 3
+
+    def effect(self, *objects) -> Iterator[cirq.Operation]:
+        source_0, target_0, target_1 = objects
```
solodov, line 350: same pattern of use of objects as above, and the type annotation

unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -217,3 +220,418 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
...
+        self.quantum_path_pieces_1 = quantum_path_pieces_1
+
+    def num_dimension(self) -> Optional[int]:
+        return 2
+
+    def num_objects(self) -> Optional[int]:
+        return 3
+
+    def effect(self, *objects) -> Iterator[cirq.Operation]:
+        source_0, source_1, target_0 = objects
```
solodov, line 441: and again here :)

unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -217,3 +220,418 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
...
+        self.quantum_path_pieces_0 = quantum_path_pieces_0
+
+    def num_dimension(self) -> Optional[int]:
+        return 2
+
+    def num_objects(self) -> Optional[int]:
+        return 2
+
+    def effect(self, *objects) -> Iterator[cirq.Operation]:
+        source_0, target_0 = objects
```
solodov, line 525: and here

unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -217,3 +220,418 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
...
+            # one quantum path piece being occupied.
+            could_capture = False
+            # We loop over all quantum path pieces and check if it could be the only
+            # occupied piece. The fire could be made if it does, otherwise not.
+            # TODO(): think a more efficient way of implementing this case.
+            for index, expect_occupied_path_piece in enumerate(quantum_path_pieces_0):
+                # TODO(): consider specific cases to save the ancilla.
+                # Add a new ancilla to indicate whether the fire could be made (value = 1 means it could).
+                capture_ancilla = world._add_ancilla(
+                    f"{expect_occupied_path_piece.name}"
```
solodov, line 597: is `expect_occupied_path_piece.name` not guaranteed to be a string?

unitary/examples/quantum_chinese_chess/chess.py
madcpf: Done. Thanks.

unitary/examples/quantum_chinese_chess/move.py
madcpf: See reply above. Thanks.

unitary/examples/quantum_chinese_chess/move.py
madcpf: See reply above. Thanks.

unitary/examples/quantum_chinese_chess/move.py
madcpf: See reply above. Thanks.

unitary/examples/quantum_chinese_chess/move.py
madcpf: Yes it is a string. It's fixed now. Thanks.

unitary/examples/quantum_chinese_chess/move.py
madcpf: This method is derived from QuantumEffect. I've updated the methods in QuantumEffects with typing annotation. Thanks.

unitary/examples/quantum_chinese_chess/move.py
```diff
@@ -217,3 +220,416 @@ def effect(self, *objects) -> Iterator[cirq.Operation]:
...
+        alpha.quantum_if(path_0_clear_ancilla, path_1_clear_ancilla).equals(1, 0).apply(
+            alpha.PhasedMove()
+        )(source_0, target_0)
+
+        # Else if only path 1 is clear, we ISWAP source_0 and target_1.
+        alpha.quantum_if(path_0_clear_ancilla, path_1_clear_ancilla).equals(0, 1).apply(
+            alpha.PhasedMove()
+        )(source_0, target_1)
+
+        # TODO(): Do we need to zero-out, i.e. reverse those ancillas?
```
dstrain115, line 405: Only if you want to reuse them.

unitary/examples/quantum_chinese_chess/chess.py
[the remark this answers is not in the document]
madcpf: Thanks Doug. I'm changing away from *. Maybe later I can changing away from importing individual classes across all files.

unitary/examples/quantum_chinese_chess/move.py
madcpf: Thanks Doug! That makes sense.

unitary/examples/quantum_chinese_chess/move.py
madcpf: Thanks Doug. QuantumEffect._verify_objects() returns the following error message now.

```
    def _verify_objects(self, *objects: "QuantumObject"):
        if self.num_objects() is not None and len(objects) != self.num_objects():
            raise ValueError(f"Cannot apply effect to {len(objects)} qubits.")
...
```

PR #173, opened 2023-12-04
[Quantum Chinese Chess] Add save_snapshot() and restore_last_snapshot() to QuantumWorld

madcpf:
- Add save_snapshot() and restore_last_snapshot() to QuantumWorld
- Apply these new methods in quantum chinese chess

### DIFF
--- a/unitary/alpha/quantum_world.py
+++ b/unitary/alpha/quantum_world.py
@@ -71,7 +71,7 @@ class QuantumWorld:
         self.effect_history: List[Tuple[cirq.Circuit, Dict[QuantumObject, int]]] = []
         # This variable is used to save the length of current effect history before each move is made,
         # so that if we later undo we know how many effects we need to pop out, since each move could
-        # consisit of several effects.
+        # consist of several effects.
         self.effect_history_length: List[int] = []
         self.object_name_dict: Dict[str, QuantumObject] = {}
         self.ancilla_names: Set[str] = set()
@@ -291,7 +291,12 @@ class QuantumWorld:
         self.circuit, self.post_selection = self.effect_history.pop()
 
     def save_snapshot(self) -> None:
-        """Saves the current length of the effect history and qubit_remapping_dict."""
+        """Saves the current length of the effect history and qubit_remapping_dict.
+
+        Normally this could default to be called after every move made by player of your
+        game, so that later if the player choose to undo his last move, we could use
+        `restore_last_snapshot` to restore the quantum properties to the snapshot.
+        """
         self.effect_history_length.append(len(self.effect_history))
         self.qubit_remapping_dict_length.append(len(self.qubit_remapping_dict))
 
--- a/unitary/alpha/quantum_world.py
+++ b/unitary/alpha/quantum_world.py
@@ -69,12 +69,22 @@ class QuantumWorld:
         """
         self.circuit = cirq.Circuit()
         self.effect_history: List[Tuple[cirq.Circuit, Dict[QuantumObject, int]]] = []
+        # This variable is used to save the length of current effect history before each move is made,
+        # so that if we later undo we know how many effects we need to pop out, since each move could
+        # consisit of several effects.
+        self.effect_history_length = []
         self.object_name_dict: Dict[str, QuantumObject] = {}
         self.ancilla_names: Set[str] = set()
         # When `compile_to_qubits` is True, this tracks the mapping of the
         # original qudits to the compiled qubits.
         self.compiled_qubits: Dict[cirq.Qid, List[cirq.Qid]] = {}
         self.post_selection: Dict[QuantumObject, int] = {}
+        # This variable is used to save the qubit remapping dictionary before each move, so that if
+        # we later undo we know how to reverse the mapping.
+        self.qubit_remapping_dict: List[Dict[cirq.Qid, cirq.Qid]] = []
+        # This variable is used to save the length of qubit_remapping_dict before each move is made,
+        # so that if we later undo we know how to remap the qubits.
+        self.qubit_remapping_dict_length = []
 
     def copy(self) -> "QuantumWorld":
         new_objects = []
@@ -96,6 +106,7 @@ class QuantumWorld:
             for circuit, post_selection in self.effect_history
         ]
         new_world.post_selection = new_post_selection
+        # TODO(): copy qubit_remapping_dict with a temp mapping.
         return new_world
 
     def add_object(self, obj: QuantumObject):
@@ -257,7 +268,8 @@ class QuantumWorld:
             self._append_op(op)
 
     def undo_last_effect(self):
-        """Restores the `QuantumWorld` to the state before the last effect.
+        """Restores the circuit and post selection table of `QuantumWorld` to the
+        state before the last effect.
 
         Note that pop() is considered to be an effect for the purposes
         of this call.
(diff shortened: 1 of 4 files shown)


unitary/alpha/quantum_world.py
```diff
@@ -69,12 +69,22 @@ def clear(self) -> None:
         """
         self.circuit = cirq.Circuit()
         self.effect_history: List[Tuple[cirq.Circuit, Dict[QuantumObject, int]]] = []
+        # This variable is used to save the length of current effect history before each move is made,
+        # so that if we later undo we know how many effects we need to pop out, since each move could
+        # consisit of several effects.
+        self.effect_history_length = []
```
solodov, line 75: should this have type annotation?

unitary/alpha/quantum_world_test.py
```diff
@@ -345,9 +346,15 @@ def test_copy(simulator, compile_to_qubits):
     assert board.circuit is not board2.circuit
     assert board.effect_history == board2.effect_history
     assert board.effect_history is not board2.effect_history
+    assert board.effect_history_length == board2.effect_history_length
+    assert board.qubit_remapping_dict_length == board2.qubit_remapping_dict_length
```
solodov, line 350: ideally these asserts should also have some sort a message (second argument to assert) to help callers debug issues

unitary/alpha/quantum_world_test.py
```diff
@@ -345,9 +346,15 @@ def test_copy(simulator, compile_to_qubits):
...
     assert board.ancilla_names == board2.ancilla_names
     assert board.ancilla_names is not board2.ancilla_names
     assert len(board2.post_selection) == 2
+    assert [key.name for key in board2.qubit_remapping_dict[-1].keys()] == [
+        "l2",
+        "ancilla_l2_0",
+    ]
```
solodov, line 357: this does look a great candidate for an assert with a message

unitary/examples/quantum_chinese_chess/chess.py
```diff
@@ -69,6 +69,9 @@ def __init__(self):
         self.game_state = GameState.CONTINUES
         self.current_player = self.board.current_player
         self.debug_level = 3
+        # This variable is used to save the classical properties of the whole board before each move is
+        # made, so that if we later undo we could recover the earlier classical state.
+        self.classical_properties_history = []
```
solodov, line 74: type annotation?

unitary/examples/quantum_chinese_chess/chess.py
```diff
@@ -476,6 +474,56 @@ def game_over(self) -> None:
...
+            or len(self.classical_properties_history) <= 1
+        ):
+            # length == 1 corresponds to the initial state, and no more undo could be made.
+            return False
+
+        # Recover the mapping of qubits to the last snapshot, remove any related post selection memory,
+        # and recover the effects up to the last snapshot (which was saved after the last move finished).
+        try:
+            world.restore_last_snapshot()
+        except:
```
solodov, line 511: is a specific exception you can capture here instead of the blanked except?

unitary/examples/quantum_chinese_chess/chess.py
```diff
@@ -476,6 +474,56 @@ def game_over(self) -> None:
...
+            return False
+
+        # Recover the classical properties of all pieces to the last snapshot.
+        self.classical_properties_history.pop()
+        snapshot = self.classical_properties_history[-1]
+        index = 0
+        for row in range(10):
+            for col in "abcdefghi":
```
solodov, line 519: you can flatten this with:

```py
for index, (row, col) in enumerate(itertools.product(range(10), "abcdefghi")):
    # do stuff
```

though this might be crossing into unreasonable python hackery territory :)

unitary/examples/quantum_chinese_chess/chess_test.py
```diff
@@ -14,6 +14,8 @@
 import pytest
 import io
 import sys
+from unitary.examples.quantum_chinese_chess.test_utils import *
```
solodov, line 17: avoid star imports?

unitary/alpha/quantum_world_test.py
madcpf: I think most cases here are kind of self explanatory and would be a bit redundant to add messages. Thanks.

unitary/alpha/quantum_world_test.py
madcpf: Added. Thanks.

unitary/examples/quantum_chinese_chess/chess.py
madcpf: Right now the possible error is only ValueError. But to be future proof I think maybe we could just include all kinds of exceptions here, instead of 

try:
   abc
except ValueError:
   return False
except:
   return False

 Thanks.

unitary/examples/quantum_chinese_chess/chess.py
madcpf: Thanks for the suggestion. I did a test trying to compare their speeds:

In [13]: s1=time.time()
    ...: for a, b in itertools.product(range(10), "abcdefghi"):
    ...:     c,d = a,b
    ...: s2 = time.time()
    ...: 
    ...: 
    ...: e1=time.time()
    ...: for row in range(10):
    ...:     for col in "abcdefghi":
    ...:         c,d = row, col
    ...: e2 = time.time()

In [14]: s2-s1
Out[14]: 0.00016570091247558594

In [15]: e2-e1
Out[15]: 0.00012350082397460938

I tried twice and double loops both wins in this case. I'll keep it like this for now. Thanks.

unitary/examples/quantum_chinese_chess/chess_test.py
madcpf: Updated. Thanks.

unitary/alpha/quantum_world.py
```diff
@@ -69,12 +69,22 @@ def clear(self) -> None:
         """
         self.circuit = cirq.Circuit()
         self.effect_history: List[Tuple[cirq.Circuit, Dict[QuantumObject, int]]] = []
+        # This variable is used to save the length of current effect history before each move is made,
+        # so that if we later undo we know how many effects we need to pop out, since each move could
+        # consisit of several effects.
```
dstrain115, line 74: nit: consist.

unitary/examples/quantum_chinese_chess/chess.py
dstrain115: I also don't think it's a good idea to catch all exceptions.  Some you might not want to catch, like "KeyboardInterrupt" for instance.  This is also going to swallow errors and hide them which will make debugging harder too.

unitary/alpha/quantum_world.py
```diff
@@ -269,6 +290,52 @@ def undo_last_effect(self):
             raise IndexError("No effects to undo")
         self.circuit, self.post_selection = self.effect_history.pop()
 
+    def save_snapshot(self) -> None:
+        """Saves the current length of the effect history and qubit_remapping_dict."""
```
dstrain115, line 294: I think we want a better description here.  Is the idea that we save a snapshot every move?

unitary/alpha/quantum_world_test.py
```diff
@@ -775,3 +782,72 @@ def test_get_correlated_histogram_with_entangled_qobjects(simulator, compile_to_
...
+    # Snapshot #2
+    world.save_snapshot()
+    # 2 more effects from Flip() and Split()
+    assert world.effect_history_length == [1, 2, 4]
+    # 2 mapping from force_measurement() and unhook()
+    assert world.qubit_remapping_dict_length == [0, 0, 2]
+    # 1 post selection from force_measurement
+    assert len(world.post_selection) == 1
+
+    # Restore to snapshot #1
```
dstrain115, line 841: Do you think it is worth testing that the measurements of the qubits/objects is correct as well?

unitary/alpha/quantum_world.py
madcpf: Fixed. Thanks.

unitary/alpha/quantum_world_test.py
madcpf: Thanks Doug. More assertions are added to test the qubit values.

dstrain115 (approved):

PR #174, opened 2023-12-15
Allow quantum if with multiple effects

dstrain115:
- Note that all effects must apply to the same qubits.

### DIFF
--- a/unitary/alpha/quantum_effect.py
+++ b/unitary/alpha/quantum_effect.py
@@ -100,7 +100,7 @@ class QuantumThen(QuantumEffect):
     def __init__(self, *objects: "QuantumObject"):
         self.control_objects = list(objects)
         self.condition = [1] * len(self.control_objects)
-        self.then_effect = None
+        self.then_effects = None
 
     def equals(
         self, *conditions: Union[enum.Enum, int, Sequence[Union[enum.Enum, int]]]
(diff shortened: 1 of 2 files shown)


unitary/alpha/quantum_effect_test.py
```diff
@@ -68,6 +68,22 @@ def test_anti_control(simulator, compile_to_qubits):
...
+    piece = alpha.QuantumObject("q0", 1)
+    piece2 = alpha.QuantumObject("q1", 0)
+    board.add_object(piece)
+    board.add_object(piece2)
+    alpha.quantum_if(piece).apply(
+        alpha.Flip(effect_fraction=0.5), alpha.Flip(effect_fraction=0.5)
+    )(piece2)
+
+    # Test that results are as expected
+    results = board.peek([piece, piece2], count=100)
```
madcpf, line 84: Is this test finished?

Also could you add another test with multiple qubits, e.g. Split?

unitary/alpha/quantum_effect.py
```diff
@@ -124,9 +124,13 @@ def then(self, effect: "QuantumEffect"):
...
         return self.apply(effect)
 
-    def apply(self, effect: "QuantumEffect"):
-        """Applies a QuantumEffect conditionally to the specified qubits."""
-        self.then_effect = effect
+    def apply(self, *effects: "QuantumEffect"):
+        """Applies a QuantumEffect conditionally to the specified qubits.
+
+        If multiple effects are specified, they will all be applied
+        (to the same qubits).
```
madcpf, line 131: Should we have a TODO() or a suggestion here for cases where effects are applied to different qubits?

PR #176, opened 2024-01-31
[Quantum Chinese Chess] Improve chess board printing

madcpf:
- Print the board with sampled probabilities.
- For proper print, different treatment is made for English vs Chinese and Sublime terminus vs mac/gLinux terminal.

### DIFF
--- a/unitary/examples/quantum_chinese_chess/board.py
+++ b/unitary/examples/quantum_chinese_chess/board.py
@@ -20,6 +20,7 @@ from unitary.examples.quantum_chinese_chess.enums import (
     Type,
     Language,
     MoveVariant,
+    TerminalType,
 )
 from unitary.examples.quantum_chinese_chess.piece import Piece
 from unitary.examples.quantum_chinese_chess.move import Jump
@@ -55,7 +56,7 @@ class Board:
         self.king_locations = king_locations
         self.lang = Language.EN  # The default language is English.
 
-    def set_language(self, lang: Language):
+    def set_language(self, lang: Language) -> None:
         self.lang = lang
 
     @classmethod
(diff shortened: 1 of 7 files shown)


madcpf:
Not sure why some other files fail Lint check.

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -28,6 +28,20 @@
 # The default initial state of the game.
 _INITIAL_FEN = "RHEAKAEHR/9/1C5C1/P1P1P1P1P/9/9/p1p1p1p1p/1c5c1/9/rheakaehr w---1"
 
+# Constants for printing board
+_RESET = "\033[0m"
```
dstrain115, line 32: Does all this work in colab?  I'm not sure how portable all these unix terminal codes are.  If it works in colab though, that's probably good enough

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -83,47 +97,157 @@ def from_fen(cls, fen: str = _INITIAL_FEN) -> "Board":
...
+    # type, e.g. sublime terminus vs glinux / mac terminal.
+    # TODO(): right now all possibilities are printed in black, maybe update to print
+    # in the same color as the corresponding pieces.
+    def to_str(
+        self,
+        probabilities: List[float] = None,
+        print_probabilities=True,
+        peek_result: List[int] = None,
+        sublime_terminus: bool = True,
+    ):
```
dstrain115, line 113: Please add docstring for the arguments.  It particular, I am not sure what sublime_terminus is?  Is it a specific terminal app?

unitary/examples/quantum_chinese_chess/board.py
```diff
@@ -83,47 +97,157 @@ def from_fen(cls, fen: str = _INITIAL_FEN) -> "Board":
...
+    # TODO(): check if there is better way to automatic determine the current terminal
+    # type, e.g. sublime terminus vs glinux / mac terminal.
+    # TODO(): right now all possibilities are printed in black, maybe update to print
+    # in the same color as the corresponding pieces.
+    def to_str(
+        self,
+        probabilities: List[float] = None,
+        print_probabilities=True,
+        peek_result: List[int] = None,
+        sublime_terminus: bool = True,
```
dstrain115, line 112: I'd consider making this an enum rather than a bool, and add a plain option, such as:

```
class TerminalType(enum.Enum):
    PLAIN_ASCII = 0
    MAC_AND_LINUX = 1
    SUBLIME_TERMINUS = 2

```

dstrain115 (commented):
Cool improvement, but let's make sure its portable!

unitary/examples/quantum_chinese_chess/board.py
madcpf: Hi Doug, yes I just checked that it works in cola, with https://colab.research.google.com/drive/1B-7LmTN_gL4T1akVqUtoYO5dwzK_3IVK?usp=sharing, but replaced the files with this PR.

unitary/examples/quantum_chinese_chess/board.py
madcpf: Thanks Doug. I just added MAC_OR_LINUX and SUBLIME_TERMINUS_OR_COLAB (since it turns out that colab handles full width chars in a similar way as sublime terminus). What do you mean about plain ascii?

PR #182, opened 2024-02-20
allow action selections to be modified

solodov:
This change adds a function to get multiple user inputs, each with its own prompt function. Once all inputs have been collected, user gets a chance to confirm selection by providing empty input (press enter) or enter "r" to select again.

### DIFF
--- a/unitary/examples/quantum_rpg/input_helpers.py
+++ b/unitary/examples/quantum_rpg/input_helpers.py
@@ -2,7 +2,7 @@
 from typing import Callable, Optional, Sequence, TextIO
 
 import sys
-import unitary.examples.quantum_rpg.qaracter as qaracter
+from unitary.examples.quantum_rpg import qaracter
 
 _USER_INPUT = Callable[[str], str]
 _INVALID_MESSAGE = "Invalid number selected."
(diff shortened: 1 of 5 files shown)


unitary/examples/quantum_rpg/input_helpers.py
```diff
@@ -54,9 +55,32 @@ def get_user_input_number(
...
+    """Runs multiple number or string prompts and returns their results.
+
+    After all inputs have been provided the last prompt asks for confirmation if
+    user happy with inputs, and allows to restart the sequence and change
+    inputs.
+    """
+    while True:
+        inputs = [p() for p in prompts]
+        print("[enter]) Confirm selection.", file=file)
+        print("r) Select again.", file=file)
```
dstrain115, line 72: Maybe we could change this to "Redo selection" so that the choice of 'r' is more obvious?

unitary/examples/quantum_rpg/input_helpers.py
```diff
@@ -54,9 +55,32 @@ def get_user_input_number(
...
+) -> Sequence[Union[int, str]]:
+    """Runs multiple number or string prompts and returns their results.
+
+    After all inputs have been provided the last prompt asks for confirmation if
+    user happy with inputs, and allows to restart the sequence and change
+    inputs.
+    """
+    while True:
+        inputs = [p() for p in prompts]
+        print("[enter]) Confirm selection.", file=file)
```
dstrain115, line 71: I am conflicted about making the user do an extra prompt at the end for every action.  Let's discuss at the next unitary meeting.

unitary/examples/quantum_rpg/input_helpers.py
solodov: done

dstrain115 (approved):
LGTM.  If you feel adventurous (pun intended), then feel free to follow up with a "fast-mode"

PR #183, opened 2024-02-27
[Quantum Chess] Improve test failure output

losos0:
See 2nd commit message for an example.

### DIFF
--- a/unitary/quantum_chess/quantum_board_test.py
+++ b/unitary/quantum_chess/quantum_board_test.py
@@ -26,7 +26,6 @@ import unitary.quantum_chess.quantum_board as qb
 from unitary.quantum_chess.test_utils import (
     assert_sample_distribution,
     assert_samples_in,
-    assert_this_or_that,
     assert_prob_about,
     assert_fifty_fifty,
 )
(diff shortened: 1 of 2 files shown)


unitary/alpha/quantum_world.py
```diff
@@ -111,9 +111,9 @@ def copy(self) -> "QuantumWorld":
         for remap in self.qubit_remapping_dict:
             new_dict = {}
             for key_obj, value_obj in remap.items():
-                new_dict[
-                    new_world.get_object_by_name(key_obj.name)
-                ] = new_world.get_object_by_name(value_obj.name)
+                new_dict[new_world.get_object_by_name(key_obj.name)] = (
+                    new_world.get_object_by_name(value_obj.name)
+                )
```
dstrain115, line 116: What's up with all these formatting changes?  Are you on the same version of black as everyone else?

dstrain115 (commented):
Test error message change looks good, but I think we should revert the formatting changes (or maybe put them in a separate PR) since they are not related to this change.

unitary/alpha/quantum_world.py
losos0: There is a CI check on this repo that fails if the formatting is not identical to that which whichever black version the workflow installs produces, so I had to add a formatting commit to make it pass.

In the mean time, Peter already hit the same issue and merged a commit making those formatting changes, so now I can remove mine.

dstrain115: ok, perfect.  I will merge this then.

Issue #178, opened 2024-02-01
Quantum RPG: Bridge Repair not in the save file


dstrain115:
When you repair the bridge, it is not in the save file.  Though, most of the time, you probably cross the bridge, so it's not a huge deal, but it does break continuity a bit.

dmasspla:
Please see [PR#186](https://github.com/quantumlib/unitary/pull/186) as an alternative to fix this issue.

The change is based on including the bridge repair in the save file, and changing the location description to guide the user towards examining the bridge to know if it's broken. Examining the bridge triggers updating the location based on the--reloaded--bridge state.

dstrain115:
Fixed by https://github.com/quantumlib/unitary/pull/186

PR #186, opened 2024-03-12
[Quantum RPG] Restore bridge repair upon examining the bridge

dmasspla:
This change hints to the user to examine the bridge to learn if it's broken. Examining the bridge effectively restores its state from previous sessions.

Prior to this change, the state of the bridge would not get persisted in the save file, and the user (with the engineer) would need to fix it whenever it loaded the game back.

This approach provides slightly better continuity, and is implemented within the constraints of the current framework.


dmasspla:
(Fixes #178)

dstrain115 (approved):
Thanks for fixing this!  This looks like the best solution we can implement right now.

PR #187, opened 2024-03-21
Use sparse simulator by default

losos0:


dstrain115 (approved):
LGTM.  Glad to see this happen.

PR #190, opened 2024-05-10
Unitary: Add generalized Hadamard gate

madcpf:
- Add generalized Hadamard gate into qudit_gates and qudit_effects
- Fix a few typos in qudit_gates and qudit_effects


unitary/alpha/qudit_effects.py
```diff
@@ -127,3 +125,27 @@ class QuditFlip(Flip):
...
+
+    def effect(self, *objects):
+        for q in objects:
+            if q.qubit.dimension == 2:
+                yield cirq.H(q.qubit)
+            else:
+                yield QuditHadamardGate(dimension=q.qubit.dimension)(q.qubit)
+
+
+class QuditSuperpose(Superpose):
```
dstrain115, line 143: For what backwards compatibility?  This is a new gate.  Let's remove this.

unitary/alpha/qudit_effects.py
```diff
@@ -127,3 +125,27 @@ class QuditFlip(Flip):
 
     def __init__(self, dimension: int, state0: int, state1: int):
         super().__init__(state0=state0, state1=state1)
+
+
+class Superpose(QuantumEffect):
```
dstrain115, line 130: Let's move this into unitary.alpha.qubit_effects.Superposition if possible.

The feedback from test users is that no one wants to worry about Qudits vs Qubits, and that we should be able to do Superpostion() and it should work on whatever dimension (which the qudit_gate you wrote already supports)

unitary/alpha/qudit_gates.py
```diff
@@ -34,28 +34,32 @@ def __init__(
...
         return (self.dimension,)
 
     def _unitary_(self):
-        arr = np.zeros((self.dimension, self.dimension))
-        arr[self.source_state, self.destination_state] = 1
-        arr[self.destination_state, self.source_state] = 1
-        for i in range(self.dimension):
-            if i != self.source_state and i != self.destination_state:
-                arr[i, i] = 1
+        arr = np.eye(self.dimension)
```
dstrain115, line 46: These are great changes, but let's split them out into a different PR since they are not relevant to the creation of a generalized Hadamard.

unitary/alpha/qudit_effects.py
madcpf: Thanks Doug. It's removed.

unitary/alpha/qudit_effects.py
madcpf: Thanks Doug. It's combined into qubit_effects.Superposition.

unitary/alpha/qudit_gates.py
madcpf: Thanks Doug. They are split out and will be shared in another PR.

dstrain115 (approved):

PR #192, opened 2024-05-22
Moved all examples out of the package directory into the top level examples directory

solodov:
Examples are no longer installed by pip when installing unitary. Each example has a README.md with instructions how to run it.

All example tests pass and each example is runnable, as far as I can tell.


solodov:
> Main thing is that relative imports are discouraged by the style guide.

Yes, this does go against the advice, however I think it's justified in this case. Absolute imports work when everything is installed as a package. However examples are not meant to be installable, they are meant to be a collection of code that runs from the examples directory, wherever it is on the filesystem. 

I assume you're talking about the decision in https://google.github.io/styleguide/pyguide.html#224-decision. The main goal of absolute imports is to avoid importing a package twice. It's not an issue in this case because example code cannot be imported with absolute imports, and unitary code cannot be imported as relative imports. I don't think the problem this decision is trying to avoid is relevant here.

dstrain115:
Fair enough.  Approved.

PR #195, opened 2024-06-26
sort exists for printing

### LINKED ISSUE #179: Quantum RPG: Order exits
The order of exits is not deterministic and maybe we could sort the exit names.

User feedback:
I've noticed sometimes the order of exits differs, e.g. "north, south" sometimes and "south, north" others.

solodov:
fixes #179


dstrain115 (approved):
Thanks for fixing this!

PR #196, opened 2024-06-27
implemented quantopedia command

### LINKED ISSUE #177: Quantum RPG: 'q' exits game when meant 'quantapedia'
From a user:
I tried the command "q" outside of battle to read the quantumpedia. Instead it quit the game without a confirmation and without saving. Another example of punishing exploration of the UI.

We should add a command to access the quantumpedia outside of battle, which would make 'q' not a unique command.

This would involve adding an entry here:
https://github.com/quantumlib/unitary/blob/main/unitary/examples/quantum_rpg/main_loop.py#L32
to list known monsters and display their quantapedia entries, ie.
https://github.com/quantumlib/unitary/blob/main/unitary/examples/quantum_rpg/npcs.py#L114

solodov:
Fixes #177

Output looks like this:

```
>qua

Observer
  Observers are known to frequent quantum events.
  They will measure qubits in order to find out their values.

BlueFoam
  Blue foam are the simplest kind of quantum errors.  Blue foam
  are usually found in the |0> state and can be measured.
  They will often slime their opponents with small fracts of X gates.

GreenFoam
  Green foam are a simple kind of quantum error.  Green foam
  are usually found in the |0> state and can be measured immediately.
  They will often ooze, which will change their opponent's phase

RedFoam
  Red foam are a slightly more dangerous type of quantum error.
  They are usually found in the |1> state and must be flipped
  before they can be safely measured.

PurpleFoam
  Purple foam are a combination of red and blue form.
  They are found in a |+> state which is a combination of
  the |0> state and |1> state.  They can be safely measured
  once a Hadamard gate has been applied.

SchrodingerCat
  Schrödinger's cat are found in a superposition of zero and one.
  This cat contains multiple qubits that are entangled so that all
  qubits are in the same state.  That is, all qubits are in a superposition
  of all ones or all zeros.  These cats have been known to apply
  Hadamard gates with their claws and measure opponents.
```

### DIFF
--- a/examples/quantum_rpg/main_loop.py
+++ b/examples/quantum_rpg/main_loop.py
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Optional, Sequence
 import enum
 import io
 import sys
+import textwrap
+from typing import List, Optional, Sequence
 
 from . import ascii_art
 from . import battle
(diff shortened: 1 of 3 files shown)


solodov:
There's an open question around ambiguous commands, quit and quantopedia clash unless input is long enough, e.g. `qua` above. With change as it is `q` will no longer quit, it will have to be `qui` or longer. I have some ideas how this can be addressed, but it would change the experience in some way. 

One idea is to make `Quit` case-sensitive, so only `Q` (and longer) would map to quit command. This is the simplest change I can think of. What do you think?

dstrain115 (approved):
Changing Quit to be capitalized seems good to me.
(in a follow-up PR though).
Thanks for the improvements!

PR #203, opened 2024-07-20
Add density matrix and entanglement calculations for qubits

madcpf:
- Add a method in quantum world to calculate density matrix of given quantum objects (only qubits are supported now);
- Based on that method, add another method in quantum world to calculate quantum mutual information of two specified qubits. The calculation is based on https://en.wikipedia.org/wiki/Quantum_mutual_information.


dstrain115:
I have some concerns about this PR.  First, this is explicitly using a different sampler than the one provided the user as part of QuantumWorld.  Second, it is using the density matrix, which is non-physical, so this will not work if the user is using a hardware device (or rather, it will use a simulator). 

Is it possible to change the partial traces in this entanglement measure to be measurements instead?  Then, this would work for both hardware and simulation, and would not rely on the density matrix or a specific simulator.

cognigami:
+Chris and Spiros for comment

My poor understanding is that this feature is meant to provide an
guesstimate of entanglement, so I'm not sure that your (otherwise spot on)
concerns matter to the output.

Chris, Spiros: thoughts?

dstrain115:
> My poor understanding is that this feature is meant to provide an guesstimate of entanglement, so I'm not sure that your (otherwise spot on) concerns matter to the output.

But we should be able to get a guesstimate of the entanglement from measurements instead of the density matrix, and, if we can do simply, we should.

cognigami:
Hi all,I am not following GitHub commits, but Doug is right that using measurements is the right way to go. Chris and I have discussed using “classical shadows” to get us the density matrix info we need for entanglement measures. I don’t remember what the final verdict was on how to implement this.Best,SpirosOn Jul 22, 2024, at 9:46 AM, Ricardo Olenewa ***@***.***> wrote:﻿+Chris and Spiros for commentMy poor understanding is that this feature is meant to provide an guesstimate of entanglement, so I'm not sure that your (otherwise spot on) concerns matter to the output. Chris, Spiros: thoughts?On Mon, Jul 22, 2024, 12:11 p.m. Doug Strain ***@***.***> wrote:
I have some concerns about this PR.  First, this is explicitly using a different sampler than the one provided the user as part of QuantumWorld.  Second, it is using the density matrix, which is non-physical, so this will not work if the user is using a hardware device (or rather, it will use a simulator).
Is it possible to change the partial traces in this entanglement measure to be measurements instead?  Then, this would work for both hardware and simulation, and would not rely on the density matrix or a specific simulator.

madcpf:
Thanks Doug and cognigami for the input! Please take a look at the updated version, which moves away from DensityMatrixSimulator.

unitary/alpha/quantum_world.py
```diff
@@ -636,8 +636,85 @@ def get_binary_probabilities(
             binary_probs.append(1 - one_probs[0])
         return binary_probs
 
+    def density_matrix(
+        self, objects: Optional[Sequence[QuantumObject]] = None, count: int = 1000
+    ) -> np.ndarray:
+        """Simulates the density matrix of the given objects.
```
dstrain115, line 642: Let's update the description here and then add some details about how we are assuming that this is a pure state so the density matrix is |psi><psi|

dstrain115 (approved):

unitary/alpha/quantum_world.py
madcpf: Done. Thanks.

PR #206, opened 2024-08-01
Allow measure_entanglement to calculate entanglement for multiple qubits

madcpf:
Update quantum world to allow measure_entanglement to calculate entanglement for multiple qubits.

### DIFF
--- a/unitary/alpha/quantum_world.py
+++ b/unitary/alpha/quantum_world.py
@@ -20,7 +20,12 @@ from unitary.alpha.quantum_object import QuantumObject
 from unitary.alpha.sparse_vector_simulator import PostSelectOperation, SparseSimulator
 from unitary.alpha.qudit_state_transform import qudit_to_qubit_unitary, num_bits
 import numpy as np
-import itertools
+from itertools import combinations
+import pandas as pd
+# from tabulate import tabulate
+# from texttable import Texttable
+# from prettytable import PrettyTable
+# from terminaltables import AsciiTable
 
 
 class QuantumWorld:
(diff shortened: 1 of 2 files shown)


unitary/alpha/quantum_world.py
```diff
@@ -20,7 +20,8 @@
 from unitary.alpha.sparse_vector_simulator import PostSelectOperation, SparseSimulator
 from unitary.alpha.qudit_state_transform import qudit_to_qubit_unitary, num_bits
 import numpy as np
-import itertools
+from itertools import combinations
```
dstrain115, line 23: Prefer `import itertools` then `itertools.combinations` below

"Use import statements for packages and modules only, not for individual types, classes, or functions."
https://google.github.io/styleguide/pyguide.html#22-imports

unitary/alpha/quantum_world.py
```diff
@@ -689,26 +690,103 @@ def density_matrix(
...
+                    # Fill in entropy [1 to num_qubit - 1]
+                    density_matrix_j = cirq.partial_trace(density_matrix_ij, [1])
+                    entropy[j] = cirq.von_neumann_entropy(
+                        density_matrix_j, validate=False
+                    )
+                entanglement[i][j] = entropy[i] + entropy[j] - entropy_pair[i][j]
+                entanglement[j][i] = entanglement[i][j]
+        names = list(self.object_name_dict.keys())
+        data_frame = pd.DataFrame(entanglement, index=names, columns=names)
+        print(data_frame.round(1))
```
dstrain115, line 788: Should we return the data_frame (or some other result) instead of printing it? 
Printing it seems generally unhelpful if we are using this within a game.

unitary/alpha/quantum_world.py
```diff
@@ -689,26 +690,103 @@ def density_matrix(
...
+        state_vector = np.array([0.0] * (2**num_qubits))
+        for key, val in histogram.items():
+            state_vector += self.__to_state_vector__(key) * np.sqrt(val * 1.0 / count)
+        density_matrix = np.outer(state_vector, state_vector)
+        reshaped_density_matrix = density_matrix.reshape((2, 2) * num_qubits)
+
+        entropy = [0.0] * num_qubits
+        entropy_pair = np.zeros((num_qubits, num_qubits))
+        entanglement = np.zeros((num_qubits, num_qubits))
+        for i in range(num_qubits - 1):
```
dstrain115, line 766: I think this block is fairly tough to follow.  I would add comments below or in the function's docstring to explain the implementation details.

PR #209, opened 2024-10-03
Allow installation to Python 3.10 environment

pavoljuhas:
This is needed for internal documentation tool.
All unit tests pass with Python 3.10.

Also restrict the CI check to use pylint-2 as
a temporary workaround for #210.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -12,8 +12,9 @@ jobs:
           python-version: '3.12'
           architecture: 'x64'
       - name: Install Pylint
+        # TODO: #210 - install the latest pylint below
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


dstrain115:
I suppose this is fine if this is blocking other doc updates, but if it is just unitary, I'd rather fix the root cause.  We pulled the copyright checker from cirq, and I think we would like to continue doing the same thing as cirq is doing.

pavoljuhas:
The pylint version downgrade here is just to let the CI pass for this PR.
The actual change here is to allow installation for 3.10 which is one
of the versions supported by the documentation tool (not sure if it can do 3.12).

dstrain115:
I see, okay.  Let's then do this for now.

dstrain115 (approved):

Issue #210, opened 2024-10-03
dev_tools/pylint fails with pylint-3 and later


pavoljuhas:
The pylint execution fails on loading the custom 'pylint_copyright_checker' plugin.
Related to https://github.com/quantumlib/Cirq/issues/6351

pavoljuhas:
There is no `pylint_copyright_checker.py` file in this repository, but there
is also no `wrong-or-nonexistent-copyright-notice` check enabled in [dev_tools/.pylintrc](https://github.com/pavoljuhas/unitary/blob/378add5650e01242f3f99f4ebf7dec26c7df7d0c/dev_tools/.pylintrc).

pylint-3 seems to error on plugins that do not exists, whereas pylint-2 silently ignored them.
It also discovers a plenty of lint issues that were not reported before (perhaps missing `--recursive=yes`).

This should be re-scoped to syncing with cirq pylint settings.

pavoljuhas:
@dstrain115 - please reopen, there are open TODO items in [dev_tools/.pylintrc](https://github.com/quantumlib/unitary/blob/b1d23aeeb7f2bcdb3eb2ccf8c93b3b24c8fe8eb2/dev_tools/.pylintrc#L9-L60) for rules that need to be fixed and enabled.

dstrain115:
@solodov If you want to help with these, we need to enable the pylint tests one-by-one and fix the resulting pylint errors.

PR #212, opened 2024-10-15
pylint - partial fix up

pavoljuhas:
- fix wildcard-import and unused-wildcard-import
- fix unused-variable
- fix unnecessary-pass and format
- fix redefined-builtin
- fix anomalous-backslash-in-string
- fix consider-using-f-string
- fix dangerous-default-value

Partially addresses #210

### DIFF
--- a/unitary/quantum_chess/mcpe_utils_test.py
+++ b/unitary/quantum_chess/mcpe_utils_test.py
@@ -57,7 +57,7 @@ def test_effect_of_swap():
 
 
 def test_distance_fn():
-    a1, a2, a3, _b1, b2, _b3 = cirq.GridQubit.rect(2, 3)
+    a1, a2, a3, _, b2, _ = cirq.GridQubit.rect(2, 3)
 
     # A gate operating on (a1, a3) will be improved by swapping a1 and a2, but
     # by how much depends on the distance function used.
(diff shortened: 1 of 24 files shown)


pavoljuhas:
@dstrain115 - PTAL

unitary/quantum_chess/mcpe_utils_test.py
```diff
@@ -57,7 +57,7 @@ def test_effect_of_swap():
 
 
 def test_distance_fn():
-    a1, a2, a3, b1, b2, b3 = cirq.GridQubit.rect(2, 3)
+    a1, a2, a3, _b1, b2, _b3 = cirq.GridQubit.rect(2, 3)
```
dstrain115, line 60: what's with underscores on some but not all of them?  Can we make it consistent?0.
same comment in unitary/quantum_chess/swap_updater_test.py

dstrain115 (approved):

unitary/quantum_chess/mcpe_utils_test.py
pavoljuhas: These were flagged by the unused-variable pylint checker.
I can change them to `_` if you think it is more readable.

dstrain115: Yeah, let's change them all to use underscores, just to be consistent.  Thanks for the cleanup!

pavoljuhas: Sure!  Done in d470f10f23d2e8d8f712cf4dce539c5a8144a3db.

PR #216, opened 2025-02-28
Add more guidance to CONTRIBUTING.md

mhucka:
This adds more general information, as well as an important warning about AI-generated code.


CONTRIBUTING.md
```diff
@@ -1,29 +1,123 @@
...
+Do your work and `git commit` your changes to your branch as needed.
+
+We use several tools to test code and perform other activities such as checking
+formatting against the style guidelines. You can run those tools locally during
+development. Wrapper scripts are located in the [`check/`](./check/)
+subdirectory to simplify running the tools.
+
+*   Run `check/pytest` to run the Pytest suite
+*   Run `check/mypy` to run the Mypy type checker
+*   Run `check/pylint` to run the Pylint code linter
```
dstrain115, line 81: I don't think we have these tools in unitary.  Either direct them to run mypy/black/pytest manually or remove the sections.

dstrain115 (approved):
Fine, but please fix the doc formatting check.

CONTRIBUTING.md
mhucka: It looks like we do have `check/pylint`, and there is a doc formatter, so I guess the best will be to keep the section and update the instructions.

mhucka:
> Fine, but please fix the doc formatting check.

Will do. I over-optimistically turned on auto-merge after filing this issue, so the changes got merged. Looking at dev_tools/, there are a couple of things that need attention there, so instead of reverting this PR, I'll fix the issues and do another PR.

PR #225, opened 2025-03-03
Add config files for misc. code/file format guidelines and tools

mhucka:
To help contributors align their work with project guidelines, this adds the following 4 configuration files:

- `.editorconfig`: [EditorConfig](https://spec.editorconfig.org) defines a file format for specifying some common coding style parameters. Many editors recognize `.editorconfig` files automatically, and there exist plugins for other editors. This particular `.editorconfig` file is the same as Cirq's and other Quantumlib projects, except for the value of `max_line_length`, which is here set to the default in [black](https://github.com/psf/black) (which is used
in unitary's CI checks) and the value used in unitary's `dev_tools/.pylintrc` file.

- `.markdownlintrc`: config file for [markdownlint](https://github.com/igorshubovych/markdownlint-cli), a popular linter for Markdown files. The settings here try to be as close to the Google style as possible while still being compatible with GitHub-flavored Markdown.

- `.hadolint.yaml`: config file for [actionlint](https://github.com/rhysd/actionlint), a linter for GitHub Actions workflows.

- `.yamllint.yaml`: config file for [yamllint](https://yamllint.readthedocs.io/en), a linter for YAML files. The config file sets the line length to the same as what's used for the `black` configuration, for consistency.


.markdownlintrc
```diff
@@ -0,0 +1,152 @@
...
+  // HTML, though it's unclear what subset PyPI supports. Google's style
+  // guide doesn't disallow using HTML, although it recommends against it. (C.f.
+  // the bottom of https://google.github.io/styleguide/docguide/style.html)
+  // It's worth noting, though, that Google's guidance has Google's internal
+  // documentation system in mind, and that system extends Markdown with
+  // constructs that make it possible to accomplish things you can't do in
+  // Markdown. Those extensions are also not available outside Google's system.
+  // Thus, although a goal of this markdownlint configuration is to match
+  // Google's style guide as closely as possible, these various factors suggest
+  // it's reasonable to relax the HTML limitation. The list below is based on
```
dstrain115, line 79: Yuck, I don't know if we want to allow HTML in markdown unless absolutely necessary.  
I don't know if it matters for a lint rc file, but I definitely don't want to encourage this.
Many of our repos get rendered via internal systems, so I think we should be stricter with this.

mhucka: OK, I dug into this a little bit more.

* It turns out Google's in-house g3doc/g3mark documentation explicitly allows HTML: "For advanced use cases, you can insert HTML directly into your Markdown file." In other words, it's discouraged, but by no means prohibited. Also, the reasons for discouraging it are not security related; rather, "Every bit of HTML hacking reduces the readability and portability of our Markdown corpus." The g3doc documentation shows examples of using HTML tags, as does the 1-page g3doc cheat sheet. The g3doc documentation has a whole section on using HTML, and lists the elements that the processor recognizes. So, it seems clear that g3doc tooling allows a subset of HTML.

* Devsite also explicitly allows HTML in Markdown files. "Markdown syntax is not meant to offer the full versatility of HTML, just the most common elements for written text. When you need to do something more complicated, you can use any inline HTML tag and block-level tags that are supported [...]" In fact, the section on images is mostly filled with uses of ``!

* Certain things can't be done (or at least, not well) without resorting to HTML in Markdown. For example, the use of `<details>` (which we use in things like Cirq's issue templates) is of course direct use of HTML and has no Markdown equivalent. Also, certain things involving setting the width of images, or alignment of blocks, can't be done without using HTML elements + attributes. As an other example (but one that we probably won't encounter much), attempting to write _about_ Markdown markup becomes very hard and ugly without being able to use `<code>` and `<pre>`. It could be debated whether these visual and layout needs are essential, but the limitations do exist. So, disallowing HTML would be problematic in at least some cases.

All that said, we could probably reduce the subset of HTML tags in this markdownlint configuration file. Perhaps we could pick what the internal Google tools allow, to ensure compatibility with that?

mhucka: I took another look at the Google Markdown guidance. It's not helpful here because  it actually allows even more HTML tags than the ones originally in this file. Still, I thought through the ones included, and deleted more, to try to reduce the subset further.

Let me know if this is any better!

mhucka: FWIW, I just noticed that Craig has been long been using raw html in his markdown files, to add link anchors. Example:

```markdown
<a class="anchor" id="build-cli"></a>
## Building fizzle as a standalone command line tool
```

dstrain115 (approved):

PR #238, opened 2025-04-15
Remove first-interactions workflow

mhucka:
The workflow has token permissions problems now that Google GitHub admins changed some of the security policies. This workflow is no longer worth the maintenance burden. Removing it from all our repos.


dstrain115 (approved):
So long first-interactions.  It was good knowing ya.

PR #239, opened 2025-04-15
Constrain pip to install NumPy < 2.0

mhucka:
The latest Cirq work with NumPy 2.0 and no longer constrains the NumPy version like it did before, but Unitary does not seem to work with NumPy 2, so we need to tell pip not to install it.


dstrain115:
This was fixed in https://github.com/quantumlib/unitary/pull/243. Running all tests in numpy 2.2.6 seems to work fine now.

PR #251, opened 2025-08-06
Fix remaining pylint issues

dstrain115:
- Fixes all pylint issues that were commented in #210 except for line-too-long.


mhucka (approved):
Tested by running pylint locally and it all looks good.

